PR #105, opened 2017-03-29
Object Detection Modular Support

aribornstein:
@nadavbar let me know what you think


src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,193 @@
...
+            // start exporting frames using the canplay eventListener
+            self.videotagging.video.oncanplay = iterateFrames;
+            self.videotagging.video.currentTime = 0;
+            self.videotagging.playingCallback();
+
+            function iterateFrames() {
+                var frameId = self.videotagging.getCurrentFrame();
+                var isLastFrame;
+
+                switch(exportUntil) {
```
nadavbar: Theoretically you can do here something more elegant instead of calling the "switch" for each frame. You can resolve the type of the "exportUntil" before you start the process and pass function that calculates the "isLastFrame". You can then use that function each time..can probably reduce the code.
You can define an object with the exportUntil types as key and functions as values and then just get the function according to that value...and init that object at the beginning of the module, etc.

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,193 @@
...
+                })
+            }
+        });
+    }
+
+    // TODO: Abstract to a module that receives a "framesReader" object as input
+    //exports frames to the selected detection algorithm format  for model training
+    this.export = function(method, exportUntil, exportPath, testSplit, cb) {
+
+        self.detectionAlgorithmManager.setExporter(method, exportPath, self.videotagging.inputtagsarray,
```
nadavbar: Why do you need the fourth argument (e.g. the number of positive frames?)..
It just the number of tagged frames BTW..

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,193 @@
...
+                                            Object.keys(self.videotagging.frames).length,
+                                            self.videotagging.video.videoWidth,
+                                            self.videotagging.video.videoHeight,
+                                            testSplit);
+
+        self.detectionAlgorithmManager.exporter.init()
+            .then(() => {
+                this.mapVideo(exportFrame,exportUntil).then(() => {
+                let notification = new Notification('Offline Video Tagger', {
+                    body: 'Successfully exported CNTK files.'
```
nadavbar: Notification says CNTK but this should be generic

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,193 @@
...
+                    });
+
+                });
+            }
+
+            //draw the frame to the canvas
+            canvasContext.drawImage(self.videotagging.video, 0, 0);
+            var data = frameCanvas.toDataURL('image/jpeg').replace(/^data:image\/\w+;base64,/, ""); // strip off the data: url prefix to get just the base64-encoded bytes http://stackoverflow.com/questions/5867534/how-to-save-canvas-data-to-file
+            var buf = new Buffer(data, 'base64');
+            var frameFileName = `${pathJS.basename(self.videotagging.src, pathJS.extname(self.videotagging.src))}_frame_${frameId}.jpg`
```
nadavbar: Please use node's path module instead of PathJS (I assume this is a leftover from a UI code)

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,193 @@
...
+            //run the model on the reviewPath directory
+            self.detectionAlgorithmManager.setReviewer(method, modelPath);
+            self.detectionAlgorithmManager.reviewer.reviewImagesFolder(reviewPath).then( (modelTags) => {
+                self.videotagging.frames = [];
+                self.videotagging.optionalTags.createTagControls(Object.keys(modelTags.classes));
+
+                //Create regions based on the provided modelTags
+                Object.keys(modelTags.frames).map( (pathId) => {
+                    var frameImage = new Image();
+                    frameImage.src = `${reviewPath}\\${pathId}`;
```
nadavbar: Use path.join, this will break on linux

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,193 @@
...
+                self.videotagging.video.currentTime = 0;
+                self.videotagging.playingCallback();
+                let notification = new Notification('Offline Video Tagger', { body: 'Model Ready For Review.' });
+                cb();
+            });
+        }
+
+        function saveFrame(frameId, fCanvas, canvasContext, saveCb){
+            canvasContext.drawImage(videotagging.video, 0, 0);
+            var writePath =  `${reviewPath}/${frameId}.jpg`
```
nadavbar: Path.join again. It won't break linux this time, but just path.join is better

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,193 @@
...
+            var data = fCanvas.toDataURL('image/jpeg').replace(/^data:image\/\w+;base64,/, ""); // strip off the data: url prefix to get just the base64-encoded bytes http://stackoverflow.com/questions/5867534/how-to-save-canvas-data-to-file
```
nadavbar: I think you use those 2 lines also in the export code..since it looks ugly with all of the base64/html Gibrish, I'll make a function out of it and have the 2 code path call it

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,193 @@
...
+            var buf = new Buffer(data, 'base64');
+            //write canvas to file and change frame
+            console.log('saving file', writePath);
+            if (!fs.existsSync(writePath)) {
+                fs.writeFileSync(writePath, buf);
```
nadavbar: Does it have to be Sync? It's not the end of the world but since you already have a callback here that you call it might be better you use the async version and call the saveCb when it finished.

src/public/js/review-configuration.js
```diff
@@ -2,20 +2,26 @@ const ipcRenderer = require('electron').ipcRenderer;
 const remote = require('electron').remote;
 const basepath = remote.app.getAppPath();
 
-
 window.onload = function(){
     document.getElementById('reviewButton').onclick = getReviewConfiguration;
     document.getElementById('cancelButton').onclick = closeWindow;
-    $('#output').val(`${basepath}/cntk`);
+    $('#output').val(`${basepath}/review`);
```
nadavbar: So..basebath is the app path, right? I wonder if it won't be better to set it as the directory that the video file is in. Since you already save the .json file next to the video file it might make more sense. Then again, very not critical.

src/lib/detection_algorithms/cntk/test/test_exporter.js
```diff
@@ -0,0 +1,19 @@
+const YoloExporter = require('../main.js').Exporter;
+const path = require('path');
+
+var exporter = new YoloExporter(path.join(__dirname, 'export_test'), ['cat', 'dog', 'mouse'], 1000, 1000);
```
nadavbar: Are you using it to test? It seems like it's just a left over from the copy&paste..probably should be removed

src/lib/detection_algorithms/main.js
```diff
@@ -0,0 +1,50 @@
...
+        return review_algorthim;
+    },
+
+
+    //Set the  exporter to the specified detection module
+    this.setExporter = function(algorithm, exportDirPath, classes, posFramesCount, frameWidth, frameHeight, testSplit) {
+         if (!Object.values(config).includes(algorithm)){
+             throw (`Error ${algorithm} module is not recognized`);
+         }
+         this.exporter = new detection_modules[algorithm].Exporter(exportDirPath, classes, posFramesCount, frameWidth, frameHeight, testSplit);
```
nadavbar: Don't export the reviewer and exporter directly - wrap them with function.
It might maker more sense to get rid of setExporter and setReviewer altogether and just have export and review functions that accept the algorithm name.

src/lib/detection_algorithms/main.js
```diff
@@ -0,0 +1,50 @@
+//read from config to find detection algorthims and paths
```
nadavbar: Consider placing this in a separate module (e.g., have a  "DetectionAlgorithmManger" folder and have this file inside) having a module that has more libraries of modules is uncommon practice in node.js..

src/lib/detection_algorithms/config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "YOLO": "yolo",
+    "CNTK": "cntk"
```
nadavbar: CNTK is not an algorithm name :)
It should be "Fast-RCNN with CNTK" or something similiar

src/lib/detection_algorithms/cntk/reviewer.js
```diff
@@ -0,0 +1,72 @@
+const path = require('path');
```
nadavbar: Rename the folder from "cntk" to cntkfastrcnn of something similiar. e.g. CNTK might support more detection algorithms.

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+// The Exporter interface - provides a mean to export the tagged frames
+// data in the expected data format of the detection algorithm
+// Constructor parameters:
+//  exportDirPath - path to the directory where the exported file will be placed
+//  classes - list of classes supported by the tagged data
+//  posFramesCount - number of positive tagged frames
+//  frameWidth - The width (in pixels) of the image frame
+//  frameHeight - The height (in pixels) of the image frame
+//  testSplit - the percent of tragged frames to reserve for test set defaults to 20%
+function Exporter(exportDirPath, classes, posFramesCount, frameWidth, frameHeight, testSplit) {
```
nadavbar: posFrameCount -> taggedFramesCount

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+                function updateIndecies(testIndecies, cb) {
+                    self.posFrameIndex = 0;
+                    self.testFrameIndecies = testIndecies;
+                    cb();
+                },
+                ensureDirExists.bind(null, self.exportDirPath),
+                ensureDirExists.bind(null, self.posDirPath),
+                ensureDirExists.bind(null, self.negDirPath),
+                ensureDirExists.bind(null, self.testDirPath)
+            ], function(err) {
```
nadavbar: function(err) -> => (err)

(If for some reason I've used it in YOLO then it should be changed too..)

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+
+         return new Promise(function(resolve, reject) {
+            async.waterfall([
+                generateTestIndecies.bind(null,self.testSplit),
+                function updateIndecies(testIndecies, cb) {
+                    self.posFrameIndex = 0;
+                    self.testFrameIndecies = testIndecies;
+                    cb();
+                },
+                ensureDirExists.bind(null, self.exportDirPath),
```
nadavbar: You can do here some very ugly javascript magic, with Map 
(e.g. [self.exportDirPath, self.posDirPath..].map(=>(item) {return ensureDirExists.bind(null, item)} )
or something similar. Very not critical though.

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+    //           coordinates are in absolute values of the image
+    //  tags - a list of objects containing the tagging data. Each object is in the format of:
+    //         {'x1' : int, 'y1' : int, 'x2' : int, 'y2' : int, 'class' : string}
+    //         Where (x1,y1) and (x2,y2) are the coordinates of the top left and bottom right corner
+    //         of the bounding boxes (respectively), and 'class' is the name of the class.
+    // Returns: A Promise object that resolves when the operation completes
+    this.exportFrame = function exportFrame(frameFileName, frameBuffer, tags) {
+        return new Promise(function(resolve, reject) {
+            async.waterfall([
+                deleteFileIfExists.bind(null, path.join(self.negDirPath,  frameFileName)),
```
nadavbar: Ok, so here I would use map.

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+                deleteFileIfExists.bind(null, path.join(self.testDirPath, frameFileName)),
+                deleteFileIfExists.bind(null, path.join(self.posDirPath,  `${frameFileName}.bboxes.labels.tsv`)),
+                deleteFileIfExists.bind(null, path.join(self.testDirPath, `${frameFileName}.bboxes.labels.tsv`)),
+                deleteFileIfExists.bind(null, path.join(self.posDirPath,  `${frameFileName}.bboxes.tsv`)),
+                deleteFileIfExists.bind(null, path.join(self.testDirPath, `${frameFileName}.bboxes.tsv`)),
+                function determineWritePath(cb){
+                    if( !tags.length){ 
+                        cb(null, self.negDirPath);
+                    } else {
+                         if (self.testFrameIndecies.includes(self.posFrameIndex)){
```
nadavbar: You can get rid of this if else by doing:

```
var dirPath = self.testFrameIndecies.includes(self.posFrameIndex) ? testDirPath : self.posDirPath;
self.posFrameIndex++;
return cb(null, self.posDirPath);
```

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+                             cb(null, self.testDirPath);
+                         } else {
+                             self.posFrameIndex++;
+                             cb(null, self.posDirPath);
+                         }
+                    }
+                },
+                function saveImage(exportPath, cb) {
+                    var imageFilePath  = path.join(exportPath, frameFileName);
+                    fs.writeFileSync(imageFilePath, frameBuffer);
```
nadavbar: You the async version instead.
You have a callback here, already..

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+                        var labelData = '';
+                        var bboxesData = '';
+                        for (var i in tags) {
+                            var tag = tags[i];
+                            labelData  += `${tag.class}\n`;
+                            bboxesData += `${tag.x1}\t${tag.y1}\t${tag.x2}\t${tag.y2}\n`;
+                        }
+                        var labelFilePath  = path.join(exportPath, `${frameFileName}.bboxes.labels.tsv`);
+                        var bboxesFilePath  = path.join(exportPath, `${frameFileName}.bboxes.tsv`);
+                        fs.writeFileSync(labelFilePath, labelData);
```
nadavbar: Use async here and below

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+       var testIndecies = [];
+       while(testIndecies.length < Math.ceil(percent * self.posFramesCount)){
+            var randomnumber = Math.ceil(Math.random() * self.posFramesCount)
+            if(testIndecies.indexOf(randomnumber) > -1) continue;
+            testIndecies[testIndecies.length] = randomnumber;
+        }
+        cb(null, testIndecies);
+    } 
+    // http://stackoverflow.com/questions/21194934/node-how-to-create-a-directory-if-doesnt-exist
+    function ensureDirExists(path, cb) {
```
nadavbar: Move it to a new utils module..
You can place it under lib/detection_algorithms
And then do require('../detectionUtil.js') both from Yolo and from here.

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+                if (err.code == 'EEXIST') { 
+                    return cb(); // ignore the error if the folder already exists
+                }
+                return cb(err);
+            }
+            cb();
+        });
+    }
+
+    function deleteFileIfExists(path, cb) {
```
nadavbar: Move to utils also

src/lib/detection_algorithms/cntk/exporter.js
```diff
@@ -0,0 +1,158 @@
...
+                    reject(err);
+                    return;
+                }
+                resolve();
+            });
+        });
+    }       
+
+    //random set http://stackoverflow.com/questions/2380019/generate-unique-random-numbers-between-1-and-100 there has to be a set way of doing this
+    function generateTestIndecies(percent, cb) {
```
nadavbar: I'd also move this to utils

nadavbar: (See comment below)

src/lib/detection_algorithms/yolo/exporter.js
```diff
@@ -0,0 +1,198 @@
...
+                    return;
+                }
+                resolve();
+            });
+        });
+    }
+
+    // code snippet used from: 
+    // http://stackoverflow.com/questions/21194934/node-how-to-create-a-directory-if-doesnt-exist
+    function ensureDirExists(path, cb) {
```
nadavbar: Update the code after you move stuff to a new utils module

src/lib/videotagging_extensions/detection.js
aribornstein: You need this to generate the test set since the exporter.exportFrame did not know whether a given frame should be a positive frame or a test frame.

src/lib/videotagging_extensions/detection.js
aribornstein: Good catch this is residual because in index.js the path module was conflicting with the path parameter of the open function.

src/lib/videotagging_extensions/detection.js
aribornstein: will do

src/lib/videotagging_extensions/detection.js
aribornstein: will do its a good convention to know

src/lib/videotagging_extensions/detection.js
aribornstein: will do

src/lib/videotagging_extensions/detection.js
aribornstein: Makes sense

src/public/js/review-configuration.js
aribornstein: that makes sense i'll do that

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: will do

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: will do i like that it will make the code look cleaner

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: i'll update that i copy and pasted your code good catch

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: yep will do 👍

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: Good catch, fixed the var names though to the following 

```
var dirPath = self.testFrameIndecies.includes(self.posFrameIndex) ? self.testDirPath : self.posDirPath;
self.posFrameIndex++;
return cb(null, dirPath);
```

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: makes sense

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: 👍

src/lib/detection_algorithms/cntk/test/test_exporter.js
aribornstein: I will update it to test the cntk version, since the inputs are the same we can use it as a sanity check

src/lib/detection_algorithms/cntk/reviewer.js
aribornstein: Understood at the time i figured CNTK was using a standard format I see now that the format is specific to this code.

src/lib/detection_algorithms/config.json
aribornstein: Understood at the time i figured CNTK was using a standard format I see now that the format is specific to this code.

src/lib/detection_algorithms/main.js
aribornstein: Got it I thought this was more organized but I will respect the conventions what you are saying makes sense

src/lib/detection_algorithms/main.js
aribornstein: That makes sense i will move them out of the exporter scope and will follow that practice.

src/lib/detection_algorithms/yolo/exporter.js
aribornstein: 👍

src/lib/detection_algorithms/main.js
aribornstein: This is actually more complicated than just deprecating setExporter and setReviewer since the exporter and reviewer are run one frame at a time. I can wrap the init and exportframe function but simply adding an export and review feature will tie the detection_algorithm_manager to the ui layer through the map video function.

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: see chat question

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: with two async writes if i pass the call back to both of them i get a race condition where one will always attempt to call the callback after write that finished first already called it. Whats the best way to handle this?

aribornstein:
@nadavbar  made 98% of suggested changes and am ready for another review

Last two changes are as follows 

- Set default outpath to the selected file directory
- Find best way for handling the async callback in the two cases I outlined

src/lib/detection_algorithms/cntk/exporter.js
nadavbar: You can just have the callback of the first trigger the second write. If you want to get fancier you can use the async module (you can use the parallel method, or forEach, or anything that work)

src/lib/detection_algorithms/cntk/exporter.js
nadavbar: Not sure what is the question, but since you already receiving a callback as a parameter, you can just do:
```
fs.writeFile(imageFilePath, frameBuffer, cb)
```
Instead of lines 91+92

aribornstein: the problem i have no way of passing the export path forward to the next function, since writeFile doesn't return it ,do you have any advice.

src/lib/detection_algorithms/cntk/exporter.js
aribornstein: i'll look up the async module. Whats the difference between using sync and nesting the callback?

aribornstein: found a work around

```js
                        var metadata = [
                            {filepath : path.join(exportPath, `${frameFileName}.bboxes.labels.tsv`), data: labelData},
                            {filepath : path.join(exportPath, `${frameFileName}.bboxes.tsv`), data: bboxesData}                            
                        ]
                        
                        async.map(metadata, (obj, cb) => {
                            fs.writeFile(obj.filepath, obj.data, cb);
                        }, cb);

```

src/lib/detection_algorithm_manager/main.js
```diff
@@ -0,0 +1,53 @@
...
+const config_path = path.join(__dirname, 'config.json')
+const config = require(config_path);
+
+//load detection modules
+var detection_modules = {}
+Object.keys(config).forEach((key) => {
+    detection_modules[key] =  require(path.join(__dirname, config[key]));
+});
+
+//removes
```
nadavbar: Probably want to remove that comment

src/lib/detection_algorithm_manager/main.js
```diff
@@ -0,0 +1,53 @@
...
+        if (!Object.values(config).includes(algorithm)){
+             throw (`Error ${algorithm} module is not recognized`);
+        }
+        this.reviewer = new detection_modules[algorithm].Reviewer(modelPath);
+        review_algorthim = algorithm;
+    }
+
+}
+
+module.exports.DetectionAlgorithmManager = DetectionAlgorithmManager;
```
nadavbar: NIT pick: it's a better practice to end with a new line IMO

src/lib/detection_algorithms/cntkfastrcnn/exporter.js
```diff
@@ -0,0 +1,121 @@
...
+    // Prepare everything for exporting (e.g. create metadata files,
+    // directories, ..)    
+    // Returns: A Promise object that resolves when the operation completes
+    this.init = function init() {
+
+         return new Promise((resolve, reject) => {
+            async.waterfall([
+                detectionUtils.generateTestIndecies.bind(null,self.testSplit, self.taggedFramesCount),
+                function updateIndecies(testIndecies, cb) {
+                    self.posFrameIndex = 0;
```
nadavbar: Why are you setting this here (posFrameIndex)? Can probably set it at the beginning of the init, no?

src/lib/detection_algorithms/cntkfastrcnn/exporter.js
```diff
@@ -0,0 +1,121 @@
...
+                    self.testFrameIndecies = testIndecies;
+                    cb();
+                },
+                async.each.bind(null,[self.exportDirPath, self.posDirPath, self.negDirPath, self.testDirPath], detectionUtils.ensureDirExists)
```
nadavbar: Very nice!

src/lib/detection_algorithms/cntkfastrcnn/exporter.js
```diff
@@ -0,0 +1,121 @@
...
+                    } else {
+                        var dirPath = self.testFrameIndecies.includes(self.posFrameIndex) ? self.testDirPath : self.posDirPath;
+                        self.posFrameIndex++;
+                        cb(null, dirPath);
+                    }
+                },
+                function saveImage(exportPath, cb) {
+                    var imageFilePath  = path.join(exportPath, frameFileName);
+                    fs.writeFile(imageFilePath, frameBuffer, () => {
+                        cb(null, exportPath);
```
nadavbar: Handle error in case the write fails, e.g. (get the error object in the callback and pass it to cb)

src/lib/detection_algorithms/cntkfastrcnn/reviewer.js
```diff
@@ -0,0 +1,72 @@
...
+	//  },
+	//  "classes": {
+	//  	    "background" : 0,
+	//  	    "human": 1,
+	//  	    "cat": 2,
+	//  	    "dog" : 3
+	//      }
+    // }
+    this.reviewImagesFolder = function reviewImagesFolder(imagesFolderPath) {
+		var model = new cntkModel.CNTKFRCNNModel({cntkModelPath : self.modelPath, cntkPath: cntkConfig.cntkPath, verbose : true});
```
nadavbar: Indentation seems to be off here (e.g. too much)

src/lib/detection_algorithms/detectionUtils.js
```diff
@@ -0,0 +1,33 @@
+//random set http://stackoverflow.com/questions/2380019/generate-unique-random-numbers-between-1-and-100 there has to be a set way of doing this
+module.exports.generateTestIndecies = function(percent, taggedFramesCount, cb) {
+    var testIndecies = [];
+    while(testIndecies.length < Math.ceil(percent * taggedFramesCount)){
```
nadavbar: Funny. this is an interview question that we used to ask.
Theoretically this code can run forever if you have a very crappy random number generator that will just generate the same numbers.
The idea is to do the random pick over a list of indices and then remove the chosen index on every iteration.
Take a look at this:
https://bost.ocks.org/mike/shuffle/

src/lib/detection_algorithms/detectionUtils.js
```diff
@@ -0,0 +1,33 @@
...
+    fs.unlink(path, (err) => {
+        if (err) {
+            if (err.code == 'ENOENT') {
+                return cb();
+            }
+            return cb(err);
+        }
+        cb();
+    });
+}
```
nadavbar: newline

src/lib/detection_algorithms/detectionUtils.js
```diff
@@ -0,0 +1,33 @@
+//random set http://stackoverflow.com/questions/2380019/generate-unique-random-numbers-between-1-and-100 there has to be a set way of doing this
+module.exports.generateTestIndecies = function(percent, taggedFramesCount, cb) {
+    var testIndecies = [];
+    while(testIndecies.length < Math.ceil(percent * taggedFramesCount)){
+        var randomnumber = Math.ceil(Math.random() * taggedFramesCount)
+        if(testIndecies.indexOf(randomnumber) > -1) continue;
+        testIndecies[testIndecies.length] = randomnumber;
+    }
+    cb(null, testIndecies);
```
nadavbar: This is a sync function, so no need to have it accept and call a callback.

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,203 @@
...
+    //exports frames to the selected detection algorithm format  for model training
+    this.export = function(method, exportUntil, exportPath, testSplit, cb) {
+
+        self.detectionAlgorithmManager.setExporter(method, exportPath, self.videotagging.inputtagsarray, 
+                                            Object.keys(self.videotagging.frames).length,
+                                            self.videotagging.video.videoWidth,
+                                            self.videotagging.video.videoHeight,
+                                            testSplit);
+
+        self.detectionAlgorithmManager.exporter.init()
```
nadavbar: Can you bundle setExporter and init together? e.g. change it to setAndInitExporter, or just .initExporter

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,203 @@
...
+                    frameExportCb(err);
+                });
+        }
+    }
+
+    
+    //allows user to review model suggestions on a video
+    this.review = function(method, modelPath, reviewPath, cb) {
+        //if the export reviewPath directory does not exist create it and export all the frames then review
+        if (!fs.existsSync(reviewPath)) {
```
nadavbar: change to exists async version if possible

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,203 @@
...
+                    reviewModel();
+                });
+            });
+        } else {
+            reviewModel();
+        }
+
+        function reviewModel() {
+            //run the model on the reviewPath directory
+            self.detectionAlgorithmManager.setReviewer(method, modelPath);
```
nadavbar: I would connect the setReviewer + reviwer.reviewImagesFolder to the same function to make the abstraction more complete..

src/lib/videotagging_extensions/detection.js
```diff
@@ -0,0 +1,203 @@
...
+                let notification = new Notification('Offline Video Tagger', { body: 'Model Ready For Review.' });
+                cb();
+            });
+        }
+
+        function saveFrame(frameId, fCanvas, canvasContext, saveCb){
+            var writePath =  path.join(reviewPath, `{frameId}.jpg`);
+            //write canvas to file and change frame
+            console.log('saving file', writePath);
+            if (!fs.exists(writePath,saveCb)) {
```
nadavbar: This is an issue - you call fs.exists which is async so the return value will always evaluate to false.
The fix is to have fs.writeFile in the callback of fs.exists

src/lib/detection_algorithm_manager/main.js
```diff
@@ -0,0 +1,53 @@
+//read from config to find detection algorthims and paths
```
nadavbar: So, my main issue with the current architecture is that it still doesn't abstract the reviewer and exporter from the rest of the code.
I also don't think that this class should stateful (e.g. have a "current" exporter and reviewer)
What I have in mind is to have methods here that accept the parameters and then go from there. One option is to have an initExporter method that will return an object with an export method in it's completion callback, or even just return an exportFrame callback that can be used by the calling code.

IMO the reviewer can just be reduced to a single "review" function that accepts the folder path.

src/lib/detection_algorithms/cntkfastrcnn/reviewer.js
aribornstein: I think it might be a github issue it renders properly in vscode
image

src/lib/detection_algorithms/cntkfastrcnn/exporter.js
aribornstein: good catch

src/lib/videotagging_extensions/detection.js
aribornstein: good catch

src/lib/detection_algorithm_manager/main.js
```diff
@@ -0,0 +1,42 @@
...
+const path = require('path');
+const config_path = path.join(__dirname, 'config.json')
```
nadavbar: Let's do the change that Limor described and get rid of the config.json file.
It should be quick -
Use fs.listdirSync (sync version is OK since this happen at require-time)  to iterate over the folder of detection algorithm.

Then, for each detection algorithm add the following to the exports at main .js:

exports.displayName (which will be like the display name at the JSON)
exports.hide (which will default to false, but you need to hide the "interface" folder)

Then change the code here to get this directly from the folder instead of the config.json

src/lib/detection_algorithm_manager/main.js
```diff
@@ -0,0 +1,42 @@
...
+         }
+         var exporter = new detection_modules[algorithm].Exporter(exportDirPath, classes, posFramesCount, frameWidth, frameHeight, testSplit);
+         exporter.init().then(()=> {
+             return cb(null, exporter.exportFrame);         
+         }).catch((err) =>{
+            return cb(err);
+         });
+    },
+
+    this.initReviewer = function (algorithm, modelPath, cb) {
```
nadavbar: You can just have a single reviewImagesFolder method here instead of initReviewer (since you call it only once), but then again I'm fine with the current interface as well since we might move at some point to "review single image"

src/lib/detection_algorithm_manager/main.js
```diff
@@ -0,0 +1,42 @@
...
+        return Object.keys(config);
+    },
+
+    //Set the  exporter to the specified detection module
+    this.initExporter = function(algorithm, exportDirPath, classes, posFramesCount, frameWidth, frameHeight, testSplit, cb) {
+         if (!Object.keys(config).includes(algorithm)){
+             throw (`Error ${algorithm} module is not recognized`);
+         }
+         var exporter = new detection_modules[algorithm].Exporter(exportDirPath, classes, posFramesCount, frameWidth, frameHeight, testSplit);
+         exporter.init().then(()=> {
```
nadavbar: You could have used a promise instead of a cb here, but I'm fine with the current implementation as well. In general it's probably better to stick to one method. I'm aware that the current APIs mix them so I don't think it's an issue.

src/lib/detection_algorithms/cntkfastrcnn/reviewer.js
nadavbar: @aribornstein  - This is probably an issue with mixing tabs and spaces. It seems that there's a mix in some of the files here. I strongly suggest fixing this (for the project) since it makes the code looks funky on different editors.

src/lib/detection_algorithms/detectionUtils.js
```diff
@@ -0,0 +1,35 @@
+//fisherYates http://stackoverflow.com/questions/2380019/generate-unique-random-numbers-between-1-and-100 
+module.exports.generateTestIndecies = function(percent, taggedFramesCount) {
+    //range call http://stackoverflow.com/questions/3895478/does-javascript-have-a-method-like-range-to-generate-an-array-based-on-suppl
+    taggedFrameIndecies = Array.apply(null, Array(taggedFramesCount)).map(function (_, i) {return i;});
+    for (i = taggedFrameIndecies.length-1; i > 1  ; i--){
```
nadavbar: Note that you don't need to go all the way down to 1, just until the portion that you need for the testIndecies

src/lib/detection_algorithms/detectionUtils.js
nadavbar: Please add a new line here

src/lib/detection_algorithms/interface/main.js
```diff
@@ -0,0 +1,2 @@
+module.exports.Exporter = require('./exporter.js').Exporter
+module.exports.Reviewer = require('./reviewer.js').Reviewer
```
nadavbar: So for the change which removes the config.json you'll have here
module.exports.hide = true;

src/lib/detection_algorithms/yolo/main.js
```diff
@@ -0,0 +1,2 @@
+module.exports.Exporter = require('./exporter.js').Exporter
```
nadavbar: So for the change which removes the config.json you'll have here 
module.exports.displayName = 'YOLO';

src/lib/detection_algorithms/cntkfastrcnn/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "main" : "main.js",
+    "dependencies": {
+        "cntk-fastrcnn": "^0.2.2"
```
nadavbar: @aribornstein  - While it make since to have this dependency here for the module, it's worth mentioning that since these modules aren't part of the node_modules directory, we won't have [npm install] running here recursively. We might want to use something like: https://github.com/emgeee/recursive-install eventually.
For now we'll just need to add an explanation at the README that you should add the modules with your extension. (which is OK)

src/lib/detection_algorithms/detectionUtils.js
aribornstein: Good catch

src/lib/detection_algorithm_manager/main.js
aribornstein: i'm going to leave this as is in case we add "review single image"

src/lib/detection_algorithm_manager/main.js
```diff
@@ -0,0 +1,41 @@
+//read from config to find detection algorthims and paths
+const path = require('path');
+const fs = require('fs');
+const detection_algorithms_path = path.join(__dirname,'../detection_algorithms');
+const detection_algorithms_dirs = fs.readdirSync(detection_algorithms_path)
```
nadavbar: I'd put this in a function, but not critical

aribornstein: Since we only use it once i'm going to leave it as is for now :)

PR #145, opened 2017-09-20
YOLO: Ensure files don't start with newlines

akx:
Sometimes, depending on chance, either train.txt or test.txt could start with a newline, which breaks Darknet, as it attempts to load a file called `''`.

So, instead of tracking whether we've started an export at all, track whether we've written anything to these files.

(To be honest, I think an exporter API that didn't use `fs.appendFile` to open-append-close the files would be better, but that would require adding a "finalize" call of some sort.)

This was reported by @moonq.


aribornstein:
Will review and merge if clean this weekend thanks so much for actively participating :)

moonq:
fyi another bug: if you export YOLO from a Video, the test.txt and
train.txt do not have newlines _at all_

aribornstein:
Okay will validate and post by the end of the week.

aribornstein:
Looks good. Merging

PR #169, opened 2018-04-25
moved all bower_components libraries into public/js folder and refactored all files to use the correct new path

itye-msft:
I suggest to move to a new branch for testing


aribornstein:
I agree

PR #173, opened 2018-04-25
Fixed: When dragging one side of tagging window, the other side moves as well…

mraxn:
Bug fix: When dragging one side of tagging window, the other side moves as well…


aribornstein:
Looks good

PR #197, opened 2018-06-16
Fix to load tags correctly when the image files have been changed in the same directory.

hermanho:
When the image files have been changed in the same directory, the saved tag will be loaded incorrectly because it is based the index of images in the directory ("videotagging.imagelist"). I have made the change that the array "videotagging.frames" will be transformed into a new array with the image filename as the key and save to a new property "framesByImagesName" in save file.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -294,6 +294,21 @@ function openPath(pathName, isDir) {
 
             if (videotagging.imagelist.length){
               videotagging.imagelist = videotagging.imagelist.map((filepath) => {return path.join(pathName,filepath)});
+              if (config && config.framesByImagesName){
+                var frames = {};
+                for (var k in config.framesByImagesName){
+                  if (config.framesByImagesName.hasOwnProperty(k)) {
+                    var fullFilepath = path.join(pathName, k);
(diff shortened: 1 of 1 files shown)


aribornstein:
Thank you for the contribution I will review and follow up over the next week or two!

aribornstein:
Can you confirm that this change won't break the export from video. I'm concerned that it might since the frames in video don't have image names

hermanho:
The new property "framesByImagesName" will not be saved in the JSON file when export form video.
https://github.com/Microsoft/VoTT/pull/197/files#diff-1fdf421c05c1140f6d71444ea2b27638R384

The new property be checked when the JSON file is loaded.
https://github.com/Microsoft/VoTT/pull/197/files#diff-1fdf421c05c1140f6d71444ea2b27638R297

It imply that the property "framesByImagesName" will not be saved in JSON when export from video and it should not be affected when opening.

The existing property "frames" is kept unchanged for backward compatibility means the export file can be opened by previous version.

aribornstein:
Amazing will validate today thank you!

PR #198, opened 2018-06-16
Fix the colors for the new tags when open from directory with saved file

hermanho:
When open from directory with saved file and input some new tags in "load-form-container", those new tags will not have colors.


aribornstein:
Thank you for the contribution  I will review and follow up. Can you update the pull request to the dev branch?

PR #207, opened 2018-07-13
Add export for KITTI format

PsychoLogicAu:
Added exporter to KITTI file format, based on the existing YOLO exporter


aribornstein:
Thank you for the contribution.

brianegge:
How do you export to KITTI format? Are there any docs?

PR #219, opened 2018-08-13
Fix KITTI export missing training images

Anthbs:
Fix for no training images being exported due to posFrameIndex only being incremented after label generation but all images are saved before the labels are generated.


aribornstein:
Will take a look and merge!

PR #261, opened 2018-10-24
Feature/import export TFRecord files

### LINKED ISSUE #262: Client should be able to import/export TFRecords
VOTT should allow the import and export of TFRecords as an additional format

PIC123:
Adds the ability to load and export TFRecord files. *When loading TFRecord, need to provide a label when opening the directory, but VOTT will load the rest of the labels from the records on read. Will resolve #262

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "dependencies": {
     "async": "^2.1.5",
     "cntk-fastercnn": "^0.1.2",
+    "crypto-js": "^3.1.9-1",
     "electron": "^1.4.1",
     "electron-window-state": "^4.0.2",
     "remote": "^0.2.6",
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "electron-window-state": "^4.0.2",
     "remote": "^0.2.6",
     "replace": "^0.3.0",
-    "send": "^0.16.2"
+    "send": "^0.16.2",
+    "tfrecord": "^0.2.0"
   }
 }
--- a/src/public/js/export-configuration.js
+++ b/src/public/js/export-configuration.js
@@ -9,6 +9,7 @@ window.onload = function(){
 ipcRenderer.on('configs', (event, configs) => {  
     $('#output').val(configs.assetFolder);
     $('#format').empty(); // remove old options
+    $('#format').append($("<option></option>").attr("value", "tfrecord").text("TF Record"));
     configs.supportedFormats.forEach( (algorithm) => {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -22,10 +22,12 @@
   "dependencies": {
     "async": "^2.1.5",
     "cntk-fastercnn": "^0.1.2",
+    "crypto-js": "^3.1.9-1",
```
pjlittle: For dependencies, I usually lock them instead of having them floating. Totally fine for now -- we can decide if it's a policy we want to adapt going forward. I haven't been burnt often by this, but it definitely happens.

src/public/js/export-configuration.js
```diff
@@ -21,8 +22,11 @@ function getExportConfiguration() {
         exportPath: $('#output').val(),
         exportFormat: $('#format').val()
     };
-
-    ipcRenderer.send('export-tags', exportConfig); 
+    if(exportConfig.exportFormat == "tfrecord") {
```
pjlittle: I try to get in the habit of doing `===` in JS -- it can help avoid some heartache/weirdness down the road. And if you're using `"tfrecord"` in more than one place, I'll create a constant for it somewhere and reference that instead of a literal.

src/public/js/video-tagging/video-tagging.html
```diff
@@ -954,10 +963,13 @@
     initImageDir: function() {
         if (!this.imagelist) return;
```
pjlittle: for consistency/readability, I'd structure conditionals like: 

```
if (blah) {

} 
```
Pretty sure this is in line w/ most linters, but I'd have to double check...

src/public/js/video-tagging/video-tagging.html
```diff
@@ -1203,29 +1244,28 @@
...
+            // Update canvas image based on new video frame
+            this.updateFrameCanvas(this.curImg);
+        }           
+        this.showAllRegions();
     },
     playImageSquence: function() {
-        if(this.sequencePlaying) {
-            this.stepFwdClicked();
-        }
+        if(this.imageIndex == this.imagelist.length - 1) this.pauseState();
```
pjlittle: ditto here

pjlittle:
I definitely have a lot more digging to do in this codebase :slightly_smiling_face: 

Just some smaller, general things, otherwise LGTM.

PIC123:
Will refactor TFRecord export logic to follow VOTT pattern

PR #324, opened 2018-11-27
added tag ad-hoc

elizabethhalper:
This feature allows the user to add a tag after the initial tag creation in the first window. Now the user can add a tag while in the middle of tagging photos.


src/public/js/video-tagging/video-tagging.html
```diff
@@ -440,6 +451,14 @@
 
             this.selectRegion(currentRegionUID)
         });
+      this.form.addEventListener("submit", function(e) {
+        e.preventDefault();
+        console.log("REACHED ADDTAG");
+        newTag = document.getElementById("newTag").value;
+        self.inputtagsarray.push(newTag);
+        console.log(self.inputtagsarray);
```
pjlittle: ditto here

PR #325, opened 2018-11-27
Remove tags on the fly

### LINKED ISSUE #226: add new label class on the fly
It would be really nice to be able to add a new class once a session has already started

PIC123:
Add ability to remove tags from tag list at bottom of app with ctrl+click. Regions with tags that are deleted become untagged.
Also changes refresh app shortcut to ctrl+R

Resolves #226


src/public/js/video-tagging/optional-tags.html
```diff
@@ -69,7 +69,17 @@
...
+                    let videotagging = document.getElementById('video-tagging');
+                    let index = videotagging.inputtagsarray.indexOf(tag);
+                    if (index !== -1){
+                        videotagging.inputtagsarray.splice(index, 1);
+                        videotagging.optionalTags.colors.splice(index, 1);
+                    }
+                    document.getElementById(tag).remove();
+                    videotagging.checkRemovedTags();
+                }
+                else if (this.tagsState[tag] == "empty" || this.tagsState[tag] == "partial") {
```
pjlittle: Update these to use `===` and super nit pick, use single quotes for strings. Good to get in the groove here, as our code linter going forward will complain :slightly_smiling_face:

PR #333, opened 2018-11-30
Fix reorder bug

PIC123:
Fixes bug caused by adding event listeners after reordering tags.


src/public/js/video-tagging/optional-tags.html
```diff
@@ -113,27 +114,35 @@
             Polymer.dom(this.$.buttonsContainer).appendChild(btn);  
         }
 
-        window.addEventListener("keyup", (e) => {
+        if(videotagging.hotkeysAssigned){
+            window.removeEventListener("keyup", this.assignHotkeys);
```
wbreza: you could avoid the use of `self = ...` if you do a `.bind(this)` on the event handler.

src/public/js/video-tagging/optional-tags.html
```diff
@@ -92,12 +94,11 @@
...
                     this.tagsState[tag] = "full";
                     self.setSelected(e.currentTarget);
-                    self.fire("ontagschanged", { selected: this.projectSelectedTags(), state: this.tagsState});
                 } else {
                     this.tagsState[tag] = "empty";
                     self.setUnselected(e.currentTarget);
-                    self.fire("ontagschanged", { selected: this.projectSelectedTags(), state: this.tagsState});
-                }         
+                } 
+                self.fire("ontagschanged", { selected: this.projectSelectedTags(), state: this.tagsState});
```
wbreza: Expected to now always fire vs. previously was only firing event in the else case.

src/public/js/video-tagging/optional-tags.html
```diff
@@ -113,27 +114,35 @@
...
+    }, 
+    assignHotkeys(e){
+        let self = videotagging.optionalTags;
             if (e.keyCode >= 49 && e.keyCode <= 57 && !e.shiftKey) {
+                // debugger;
                 let index = e.keyCode - 49;
-                if (index < this.tags.length) {
-                    let tag = this.tags[index];
-                    let btn = this.getBtnByTag(tag);
+                if (index < videotagging.optionalTags.tags.length) {
```
wbreza: Why the change to use `optionalTags`.  or was `this` previously referring to `videotagging.optionTags` ?

pjlittle:
Nice to see this working.

src/public/js/video-tagging/optional-tags.html
PIC123: yah, in the context of the event listener callback we wanted the this to refer to optional tags

PR #336, opened 2018-11-30
Bug/stop propogation

elizabethhalper:
This PR addresses the "add tag" feature and suppresses all hotkeys while the tag text box is focused. It re-enables the hotkeys as soon as a new tag is submitted


kichinsky:
Hi, I made  some changes to make it mergable, and with those changes I think you don't need any more to disable/enable hotkeys on toolbar and freeze regionsManager (I added some filters in eventlisteners to ignore keyboard events on Input/Textarea/Select elements.

Can you check if it works without those calls?

elizabethhalper:
@kichinsky it does work. I merged in your changes and have the final PR ready to go. Thanks!

PR #341, opened 2018-12-04
Switching to the new CanvasTools.Editor

kichinsky:
CT Updates:
* New Editor class to wrap editor creation from separate components, moved SVG, Canvas and Filters management into CT.Editor.
* Changes to AreaSelector and RegionsManager interact on region creation/manipulation (temporary freezing is used now to simplify region creation).
* Styling updates to regions.
* Added PointRegion for RegionsManager (not integrated yet with VoTT code). + Code refactoring to allow future extensibility with new regions types.
* Updates mouse wheel support for template-tool (**wheel** and **shift+wheel**).
* Cleaned code, removed unnecessary elements.
* Autoresizing of Editor.


kichinsky:
@elizabethhalper, @PIC123 I added in the latest commit an option to mark tags for auto applying (alt+click or alt+number). Could you please check that it does not break your changes on adding/removing tags, etc.?

@aribornstein Is it something you asked for in #344?

Issue #298, opened 2018-11-14
Visually differentiate release and dev versions of app


kichinsky:
In addition to #296 -- it would be great to have some visual metaphor (e.g. different color of app icon) to clearly distinguish the version of app.

Issue #296, opened 2018-11-14
Add VoTT version/branch to title bar of the app


kichinsky:
It would be great if we could display app version in the title bar to simplify version recognition from the screenshot.

kichinsky:
Resolved in #420

PR #351, opened 2018-12-10
Add the version number to the help menu

### LINKED ISSUE #347: Add version number somewhere in the app
I would like to be able to see what version of VoTT I am using somewhere in the app. Specifically, I am talking about the release version that I downloaded and installed.  Like a little "About" menu item that many software have.

gboeer:
PR related to #347, #298, #297, #296 to supply information about the currently used VOTT version. This way, the version number will simply be displayed as entry inside the "Help" menu. It may be a small stepp to add it to the titlebar or maybe as color coded information as mentioned in those issues.

Note, this is the version as defined in the package.json.
https://github.com/Microsoft/VoTT/blob/ec6057c4c95780f7547d5c55245c6f48b396e29c/package.json#L3

This works when starting the app with npm start, however may fail with a build version (not tested).

close #347

### DIFF
--- a/main.js
+++ b/main.js
@@ -6,6 +6,8 @@ const BrowserWindow = electron.BrowserWindow;
 const windowStateKeeper = require('electron-window-state');
 const path = require('path');
 const url = require('url');
+// To access the version defined in package.json
+require('pkginfo')(module, 'version');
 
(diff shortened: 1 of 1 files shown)


Legor:
Thanks, although this touches the issues mentioned above this truly only closes #347

PR #354, opened 2018-12-10
Retrieving Assets from Azure Blob Storage in editor page

tbarlow12:


src/providers/storage/azureBlobStorage.ts
```diff
@@ -138,13 +141,41 @@ export class AzureCloudStorageService implements IStorageProvider {
             path = path ? [this.options.containerName, path].join("/") : this.options.containerName;
         }
         const files = await this.listFiles(path);
-        return files.map((blobPath) => AssetService.createAssetFromFilePath(blobPath));
+        const result: IAsset[] = [];
+        for (const key of Object.keys(files.entries)) {
+            const url = this.getUrl(files.entries[key].name);
+            const asset = AssetService.createAssetFromFilePath(url);
+            if (asset.type === AssetType.Image || asset.type === AssetType.Video) {
```
wbreza: I would consider inverting this to check whether or not it is unknown which is the 3rd option in that enum.

src/providers/storage/azureBlobStorage.ts
```diff
@@ -138,13 +141,41 @@ export class AzureCloudStorageService implements IStorageProvider {
...
+            }
+        }
+        return result;
     }
 
     private getService() {
         return AzureStorageBlob.createBlobService(this.options.connectionString);
     }
 
+    private getUrl(blobName: string) {
```
wbreza: Is this a static call via the library? Or does it make an HTTP request to figure it out?

src/providers/storage/azureBlobStorage.ts
```diff
@@ -138,13 +141,41 @@ export class AzureCloudStorageService implements IStorageProvider {
...
+            this.getHostName(this.options.connectionString),
+        );
+    }
+
+    private getHostName(connectionString: string): string {
+        const accountName = this.getAccountName(connectionString);
+        return `https://${accountName}.blob.core.windows.net`;
+    }
+
+    private getAccountName(connectionString: string): string {
```
wbreza: This method can be static

src/providers/storage/azureBlobStorage.ts
```diff
@@ -138,13 +141,41 @@ export class AzureCloudStorageService implements IStorageProvider {
             path = path ? [this.options.containerName, path].join("/") : this.options.containerName;
         }
         const files = await this.listFiles(path);
-        return files.map((blobPath) => AssetService.createAssetFromFilePath(blobPath));
+        const result: IAsset[] = [];
```
wbreza: Any tests needed to get created or updated for these changes?

src/providers/storage/azureBlobStorage.ts
```diff
@@ -1,12 +1,14 @@
...
 import AzureStorageBlob from "../../vendor/azurestoragejs/azure-storage.blob.js";
-import { IAsset } from "../../models/applicationState";
+import { IAsset, AssetType } from "../../models/applicationState";
 import { AssetService } from "../../services/assetService";
 
 export interface IAzureCloudStorageOptions {
     connectionString: string;
     containerName: string;
     createContainer: boolean;
+    accountName?: string;
```
wbreza: Doesn't look like you are using these outside of the options object here.

src/providers/storage/azureBlobStorage.ts
tbarlow12: Done

src/providers/storage/azureBlobStorage.ts
tbarlow12: I believe it's a static call

src/providers/storage/azureBlobStorage.ts
tbarlow12: I've been playing around with this, I think I'll add them once the new stuff is in. That way it'll more accurately reflect the new data structure

tbarlow12: nvm, almost done adding tests now

PR #356, opened 2018-12-10
Fix issue with different color icons on homepage

wbreza:
- The open folder didn't have an href attribute defined which caused the browser to not render it with the same styles as the other link.


wbreza:
Without the href an anchor tag basically renders like a span


________________________________

PR #357, opened 2018-12-11
Fixed tsc compile error on editor footer component

wbreza:


wbreza:
@tbarlow12 - tsc compile was failing.  Not sure why this wasn't caught on build checks but we should fix that.

PR #363, opened 2018-12-12
Project Settings Tests & Mock Factory

tbarlow12:


src/models/mockFactory.ts
```diff
@@ -0,0 +1,200 @@
...
+                color: "#FFFFFF",
+            },
+            {
+                name: "Tag2",
+                color: "#FFFF00",
+            },
+        ];
+    }
+
+    public connection(): IConnection {
```
wbreza: Any reason for all of these methods not to be static?

src/models/mockFactory.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    public exportFormat(): IExportFormat {
+        return {
+            providerType: "Fake",
+            providerOptions: {},
+        };
+    }
+
+    public project(): IProject {
```
wbreza: There is some overlap here vs. what i recently introduced - Let's consolidate

src/react/components/common/tagsInput/tagsInput.tsx
```diff
@@ -156,6 +156,12 @@ export default class TagsInput extends React.Component<ITagsInputProps, ITagsInp
         tag.text = this.ReactTagHtml(tag.id, tag.color);
     }
 
+    private getReactTags(props): IReactTag[] {
+        const tags = (props.tags) ? props.tags : props.value;
```
wbreza: Interesting - so I'm assuming in the context of the form the tags are passed by the `value` prop, but in the editor page tags are passing in the `tags` props.  This introduces some complexity.  Let's chat.

src/react/components/pages/projectSettings/projectForm.json
```diff
@@ -21,5 +21,6 @@
             "title": "Tags",
             "type": "string"
         }
-    }
+    },
+    "required": ["name", "sourceConnectionId", "targetConnectionId"]
```
wbreza: Thanks for adding this - Not sure how I missed it earlier.

src/react/components/pages/projectSettings/projectForm.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+import { BrowserRouter as Router } from "react-router-dom";
+import { IApplicationState } from "../../../../models/applicationState";
+import initialState from "../../../../redux/store/initialState";
+import createReduxStore from "../../../../redux/store/store";
+import ProjectSettingsPage from "./projectSettingsPage";
+
+describe("Project Form Component", () => {
+    let onSubmit: (value: any) => void;
+
+    const mockFactory = new MockFactory();
```
wbreza: If you convert your methods to static you wouldn't need to create an instance... I don't believe you are tracking any state within the object, correct?

src/react/components/pages/projectSettings/projectForm.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+                            project={project}
+                            connections={connections}
+                            onSubmit={onSubmit}
+                            />
+                    </Router>
+                </Provider>,
+            ).find(ProjectForm).childAt(0);
+        });
+
+        it("should have name loaded correctly", () => {
```
wbreza: You could likely combine these into a single test called `Should initialize state correctly` and then have all of your expectations there.  The initialization of these tests are all the same and you are causing the tests to run more slowly having to reinstantiate the wrapper/component for each test.

src/react/components/pages/projectSettings/projectForm.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+                        <ProjectForm
+                            project={null}
+                            connections={connections}
+                            onSubmit={onSubmit}/>
+                    </Router>
+                </Provider>,
+            ).find(ProjectForm).childAt(0);
+        });
+
+        it("should have name loaded correctly", () => {
```
wbreza: Same here - combine some tests to reduce component bootstrapping.

src/react/components/pages/projectSettings/projectForm.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+    const defaultState: IApplicationState = initialState;
+    const store = createReduxStore(defaultState);
+
+    describe("Existing project", () => {
+        onSubmit = jest.fn();
+        let wrapper: any = null;
+
+        beforeEach(() => {
+            wrapper = mount(
+                <Provider store={store}>
```
wbreza: I don't believe the project form component itself has any direct dependencies on redux or router and you can likely remove these wrapping components.

src/models/mockFactory.ts
tbarlow12: Will do. I'll make them all static and put them in your class

src/react/components/common/tagsInput/tagsInput.tsx
tbarlow12: Exactly. This was part of the reason the tests were failing when trying to mount the project form.

src/react/components/pages/projectSettings/projectForm.test.tsx
tbarlow12: Correct. I'll move all the mocks to your class and make them static

src/react/components/pages/projectSettings/projectForm.test.tsx
tbarlow12: makes sense. I'll make that change

src/react/components/pages/projectSettings/projectForm.test.tsx
tbarlow12: I got some interesting errors when I tried to run without it. Something about a `Link`. But I'll remove and get back to you with a more specific error message

src/react/components/common/tagsInput/tagsInput.tsx
tbarlow12: And I don't know for sure if it's the editor page that passes it in this way. This was mainly an artifact of testing. Maybe if I remove it from `Provider` and `Router` this won't happen?

src/react/components/pages/projectSettings/projectForm.test.tsx
tbarlow12: Update: Here's the error message I get: `    Invariant Violation: You should not use <Link> outside a <Router>`

tbarlow12: I'm guessing it's from the `Add Connection` button

src/react/components/common/tagsInput/tagsInput.tsx
tbarlow12: I figured this out. Because of the custom widget, `props.value` was being used as the default for passing the tags data to the component. By augmenting the existing `createUiSchema`, I was able to actually instantiate the `TagsInput` component and specify the `onChange` function and the `tags` data

src/react/components/pages/projectSettings/projectForm.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+describe("Project Form Component", () => {
+
+    const project = MockFactory.project();
+    const connections = MockFactory.connections();
+    const defaultState: IApplicationState = initialState;
+    const store = createReduxStore(defaultState);
+
+    function createComponent(props: IProjectFormProps) {
+        return mount(
+            <Provider store={store}>
```
wbreza: Should be safe to remove Provider wrapping tag

PR #381, opened 2018-12-14
Fixed tsc compile errors from JSON files

wbreza:
- changed JSON file imports to use require
- updated build definitions to run tsc as part of build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ before_script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sleep 3; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then npm install; fi
 script:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then tsc; fi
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -22,6 +22,7 @@ before_script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sleep 3; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then npm install; fi
 script:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then npm run compile; fi
```
mydiemho: what does compile do?  Why are we adding it now?

tbarlow12: @mydiemho this checks to make sure it will compile with typescript. it should have been added from the start

wbreza: @mydiemho It basically runs the typescript compiler.  We have see cases where lint and tests pass but there is a minor TS issue that prevents the electron app from launching.  This will run the TS compiler `tsc` in addition to linting to verify that this doesn't happen.

PR #382, opened 2018-12-14
Added unit tests for project service

wbreza:


src/services/projectService.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    it("Save throws error if storage provider cannot be created", async () => {
+        const expectedError = new Error("Error creating storage provider");
+        const createMock = StorageProviderFactory.create as jest.Mock;
+        createMock.mockImplementationOnce(() => { throw expectedError; });
+
+        await expect(projectSerivce.save(testProject)).rejects.toEqual(expectedError);
+    });
+
+    it("Delete calls project storage proivder to delete project", async () => {
```
mydiemho: small spelling error.

`proivder` should be `provider`

src/services/projectService.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        const expectedError = "Error deleting from storage provider";
+        storageProviderMock.deleteFile
+            .mockImplementationOnce(() => Promise.reject(expectedError));
+
+        await expect(projectSerivce.delete(testProject)).rejects.toEqual(expectedError);
+    });
+
+    it("Delete call fails if storage provider cannot be created", async () => {
+        const expectedError = new Error("Error creating storage provider");
+        const createMock = StorageProviderFactory.create as jest.Mock;
```
mydiemho: Is there a difference between creating the mock here vs up above like deleteFile?

PR #387, opened 2018-12-15
Editor Toolbar components

wbreza:
- ToolbarItemFactory -> to register toolbar items
- EditorToolbar -> top level toolbar component
- ToolbarItem -> abstract component with basic rendering
- Stubbed toolbar commands
- Unit Tests


src/react/components/toolbar/toolbarItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    public render() {
+        const className = ["toolbar-btn"];
+        if (this.props.active) {
+            className.push("active");
+        }
+
+        return (
+            <button type="button"
+                className={className.join(" ")}
```
JacopoMangiavacchi: Are we intentionally want to apply two css class here for the active button ?  I'm newbie here so this is just to understand but it shoudn;t be better to have a specific toolbar-btn-active class ?

wbreza: Yes, this in intentional, CSS is all about layering (cascading) multiple classes to achieve the desired results, otherwise you would end up with many duplicated styles across several different different CSS selectors.

tbarlow12:
@wbreza looks like it's just a `tsc` compiler error, but the tests pass. Looks good

wbreza:
@tbarlow12 - Fixed `tsc` compile error.

PR #391, opened 2018-12-17
add npm script for building electron app

mydiemho:


package.json
```diff
@@ -49,18 +66,21 @@
...
     "@types/react-dom": "16.0.9",
+    "@types/react-router-dom": "^4.3.1",
     "@types/redux-logger": "^3.0.6",
     "@types/redux-thunk": "^2.1.0",
+    "electron-builder": "^20.38.3",
     "enzyme": "^3.7.0",
     "enzyme-adapter-react-16": "^1.7.0",
     "foreman": "^3.0.1",
     "jest-enzyme": "^7.0.1",
+    "jquery": "^3.3.1",
```
wbreza: jquery and popper are not required in v2.

package.json
```diff
@@ -2,12 +2,11 @@
   "name": "vott-react-typescript",
   "version": "0.1.0",
   "private": true,
-  "main": "build/electron/bundle.js",
+  "main": "src/electron/main.ts",
```
wbreza: This causes the app startup to fail for devs running `npm start`.  We need an electron bundle to be used as the start script.

package.json
```diff
@@ -2,12 +2,11 @@
...
   "dependencies": {
     "axios": "^0.18.0",
     "bootstrap": "^4.1.3",
     "deepmerge": "^2.2.1",
-    "electron": "^3.0.10",
```
wbreza: We are using some electron API's at runtime - she will cause it to fail.

wbreza:
@mydiemho - current builds are also failing.

PR #392, opened 2018-12-17
Redux actions unit tests

wbreza:
- Application actions
- Connectivity actions
- Project actions


src/common/mockFactory.ts
```diff
@@ -56,7 +65,7 @@ export default class MockFactory {
             id: `project-${name}`,
             name: `Project ${name}`,
             assets: {},
-            exportFormat: null,
+            exportFormat: MockFactory.exportFormat(),
```
tbarlow12: More of a question than anything. I've seen people still use `this` while using another static function from the same class. Which way is best?

wbreza: Technically it shouldn't work since there is no instance of the class for this to work against.  Since all these methods are static we should use it directly off the `MockFactory` like in this example.

PR #395, opened 2018-12-18
Project Form Bug Fix & Increased Test Coverage

tbarlow12:


src/react/components/pages/projectSettings/projectForm.test.tsx
```diff
@@ -18,167 +19,204 @@ describe("Project Form Component", () => {
...
+    });
+
+    it("empty project has initial state loaded correctly", () => {
+        const onSubmit = jest.fn();
+        const wrapper = createComponent({
+            project: null,
+            connections,
+            onSubmit,
         });
+        expect(wrapper.state().formData.name).toBe(undefined);
```
wbreza: Nit: May be worth setting a variable to `wrapper.state()` so you don't have to call it within each expectation.

src/react/components/pages/projectSettings/projectForm.ui.json
```diff
@@ -9,6 +9,6 @@
         "ui:widget": "connectionPicker"
     },
     "tags": {
-        "ui:widget": "tagsInput"
+        "ui:field": "tagsInput"
```
wbreza: Are you seeing that the outer field template wrapper isn't rendering with the control? Like the label, etc?

PR #396, opened 2018-12-18
Add JSDocs and Cleanup TagsInput, Modal, ProjectForm, Tests

tbarlow12:


src/react/components/common/tagsInput/tagsInput.test.tsx
```diff
@@ -50,29 +47,42 @@ describe("Tags Input Component", () => {
...
-        wrapper.find("input").simulate("keyDown", {keyCode: 13}); // enter
+        wrapper.find("input").simulate("keyDown", {keyCode: KeyCodes.enter}); // enter
+        expect(onChangeHandler).toBeCalled();
+        expect(wrapper.state().tags).toHaveLength(originalTags.length + 1);
+        const newTagIndex = originalTags.length;
+        expect(wrapper.state().tags[newTagIndex].id).toEqual(newTagName);
+        expect(TagColors).toContain(wrapper.state().tags[newTagIndex].color);
+    });
+
+    it("create a new tag from text box - enter key", () => {
```
PIC123: Is this a repeat of the above test?

src/react/components/common/tagsInput/tagsInput.tsx
```diff
@@ -9,30 +9,58 @@ import { ITag } from "../../../../models/applicationState";
...
     enter: 13,
     backspace: 8,
 };
 
+/**
+ * Keys that, when pressed, cause creation of new tag
+ */
 const delimiters = [KeyCodes.comma, KeyCodes.enter];
 
 export default class TagsInput extends React.Component<ITagsInputProps, ITagsInputState> {
```
elizabethhalper: Super nit-picky, but I think this reads easier when the following are in the same order that they are when they're defined later on in the code

tbarlow12: Agreed :) I'll make the fix

src/react/components/common/tagsInput/tagsInput.test.tsx
tbarlow12: Definitely was. Ctrl+V too many times :)

PR #397, opened 2018-12-19
ConnectionPage & ConnectionForm Tests

JordanEllis6809:


JordanEllis6809:
Build breaks only on linux, looks like enzyme is behaving differently... going to test this on linux now.

JordanEllis6809:
I've found that there are 2 tests that are causing all of the others that come after them to fail. If you comment out tests:
it("adds connection when submit button is hit"
and
it("removes unselected connection when delete button is hit"

I hypothesize this is due to async background tasks that are still on the queue, using setImmediate(()... still doesn't fix the issue. A bit unsure how to proceed, will continue looking.

JordanEllis6809:
Additionally, running each test individually works just fine.

JordanEllis6809:
https://github.com/Microsoft/VoTT/issues/407

PR #401, opened 2018-12-19
Started basic localization implementation

tbarlow12:
This is not a full implementation of localization. Started with very basic pages: home page and connections page. One thing that is different here is creating the `formSchema` inline because the values will be dynamic depending on the language


src/common/strings.ts
```diff
@@ -0,0 +1,183 @@
...
+        description: string;
+        submit: string;
+        cancel: string;
+    };
+    homePage: {
+        newProject: string;
+        openProject: string;
+        recentProjects: string;
+    };
+    menuBar: {
```
wbreza: FYI - these won't show in the real app.

src/react/components/pages/connections/connectionForm.tsx
```diff
@@ -2,11 +2,47 @@ import React from "react";
 import Form, { Widget } from "react-jsonschema-form";
 import { IConnection } from "../../../../models/applicationState.js";
 import LocalFolderPicker from "../../common/localFolderPicker";
-// tslint:disable-next-line:no-var-requires
-const formSchema = require("./connectionForm.json");
+import { strings } from "../../../../common/strings";
 // tslint:disable-next-line:no-var-requires
 const uiSchema = require("./connectionForm.ui.json");
 
+const formSchema = {
```
wbreza: I'd like to see if we can somehow references the localized strings from within the JSON in some sort of variables notation like `${strings.common.displayName}`

wbreza: Then we can run the object through some parser that does some deep scanning / replacements?

src/common/strings.ts
```diff
@@ -0,0 +1,183 @@
...
+            pan: string;
+            drawRectangle: string;
+            drawPolygon: string;
+            saveProject: string;
+            exportProject: string;
+        }
+    };
+}
+
+export const strings: IStrings = new LocalizedStrings({
```
wbreza: Can these language specific strings somehow be save as json files with some naming convention like `strings.en.json` and `strings.es.json` ?

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -49,13 +50,13 @@ export default class HomePage extends React.Component<IHomepageProps> {
                         <li>
                             <Link to={"/projects/create"} className="p-5">
                                 <i className="fas fa-folder-plus fa-9x"></i>
-                                <h6>New Project</h6>
+                                <h6>{strings.homePage.newProject}</h6>
```
wbreza: Nice and easy syntax :)

PR #404, opened 2018-12-19
[AB#16328] Fixes issues with running app from static files

wbreza:
- Updated electron bundle to bundle as `main.js` into the `/build` folder.
- Moved azure-pipelines to root
- Updated `electron/main.ts` to use loadFile vs loadUrl on prod vs. dev builds


docs/PACKAGING.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Command
+
+```bash
+npm run release
+```
+
+### Underneath the hood
+
+1. Create the `react` bundle inside the `build` directory,
```
tbarlow12: Did you mean to do `1.` for all list items? Or a bullet point?

mydiemho: Markdown automatically convert the 1 to the right count for list :D.

package.json
```diff
@@ -58,11 +61,15 @@
     "@types/redux-logger": "^3.0.6",
     "@types/redux-mock-store": "^1.0.0",
     "@types/redux-thunk": "^2.1.0",
+    "electron": "^3.0.13",
+    "electron-builder": "^20.38.3",
     "enzyme": "^3.7.0",
     "enzyme-adapter-react-16": "^1.7.0",
     "foreman": "^3.0.1",
     "jest-enzyme": "^7.0.1",
+    "jquery": "^3.3.1",
```
wbreza: We're currently not using jQuery or popper.  I know they are required if you are using the bootstrap JS, but we aren't right now.

mydiemho: just for note keeping sake, electron-builder won't build without these 2 (probably some strict requirement that all peer dependencies must be install)

PR #412, opened 2018-12-20
[AB#16247] pipeline for building release artifact

mydiemho:


release-pipeline.yml
```diff
@@ -0,0 +1,34 @@
+trigger: 
+  - v2 # disable CI build for branch except v2
```
tbarlow12: Is there a way to just trigger on tags?

release-pipeline.yml
```diff
@@ -0,0 +1,34 @@
+trigger:
```
wbreza: Can we also add a trigger to only build if we push tags on v2 ?

release-pipeline.yml
mydiemho: I'll try some thing to see if it works.  Right now it will kick off a build that build the artifacts, and the release will run, but when there's no tag, it just won't do anything.  

See example:
https://dev.azure.com/msft-vott/VoTT/_releaseProgress?_a=release-environment-logs&releaseId=1&environmentId=1

release-pipeline.yml
mydiemho: I'll try some thing to see if it works.  Right now it will kick off a build that build the artifacts, and the release will run, but when there's no tag, it just won't do anything.  

See example:
https://dev.azure.com/msft-vott/VoTT/_releaseProgress?_a=release-environment-logs&releaseId=1&environmentId=1

PR #414, opened 2018-12-21
Tag Editor Modal Fix & Tests

tbarlow12:


src/react/components/common/tagsInput/tagEditorModal/tagEditorModal.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    it("modal is initialized properly", () => {
+        const onCancel = jest.fn();
+        const onOk = jest.fn();
+        const wrapper = createComponent({
+            tag,
+            showModal: false,
+            onCancel,
+            onOk,
+        });
```
elizabethhalper: Is it possible to do these few lines in a beforeEach to reduce code redundancy in the tests below?

tbarlow12: So that's what I had initially, but I was hitting some race conditions with some assertions. For example, the `onOk` handler is called with different tags in different tests. But I can probably move the `onCancel` and a default `wrapper` to the beforeEach :)

PR #416, opened 2018-12-21
Added basic confirmation component

wbreza:
- Added confirm component
- Integrated confirmation in homepage and connections listing

Test pending

### DIFF
--- a/src/react/components/pages/homepage/homePage.test.tsx
+++ b/src/react/components/pages/homepage/homePage.test.tsx
@@ -17,6 +17,10 @@ jest.mock("../../../../services/projectService");
 import ProjectService from "../../../../services/projectService";
 
 describe("Connection Picker Component", () => {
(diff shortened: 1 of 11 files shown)


src/react/components/pages/homepage/homePage.test.tsx
```diff
@@ -2,91 +2,138 @@ import React from "react";
...
                 </Router>
             </Provider>,
         );
-    });
+    }
 
     it("should render a New Project Link", () => {
+        const store = createStore(recentProjects);
+        const props = createProps();
+        const wrapper = createComponent(store, props);
```
elizabethhalper: Is there a reason you are not putting these in a beforeEach method?

wbreza: I can likely move them into a `beforeEach` tag

PR #417, opened 2018-12-21
v2 Editor Page

PIC123:
Adds CanvasTools integration into the editor page and exposes methods for use in the app 
Things that still need to be addressd:

- Currently generates random tags, need to interface with Footer/Tag component
- Needs tests
- Fix canvas resizing issue when switching to images of different sizes when the editor is small (ok when editor is larger than image)


src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+    constructor(props, context) {
+        super(props, context);
+
+        this.state = {
+            loaded: false,
+        };
+    }
+
+    public componentDidMount() {
```
tbarlow12: Seems like this could use a bit of cleanup. Maybe add some functions that are called by this one (`getTags`, `onRegionMove`, `onRegionDelete`, etc.) This function is just big, which makes it a little more challenging to read

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,329 @@
...
+            this.updateEditor();
+        }
+    }
+
+    public render() {
+        // const { loaded } = this.state;
+        // const { svgHost } = this.props;
+
+        return (
+            <div id="ctZone">
```
tbarlow12: Should this contain the CanvasTools component? Has that been reactified yet?

PIC123: So currently this is how canvas tools is initialized. It uses those id names to generate the editor.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,366 @@
...
+        this.selectRegionById = this.editor.RM.selectRegionById.bind(this.editor.RM);
+        this.toggleFreezeMode = this.editor.RM.toggleFreezeMode.bind(this.editor.RM);
+        this.unfreeze = this.editor.RM.unfreeze.bind(this.editor.RM);
+        this.updateTagsById = this.editor.RM.updateTagsById.bind(this.editor.RM);
+        this.updateTagsForSelectedRegions = this.editor.RM.updateTagsForSelectedRegions.bind(this.editor.RM);
+
+        this.incrementalRegionID = 100;
+
+        // Set up random tags for regions (will replace with footer interaction)
+        this.primaryTag = new ct.Core.Tag(
```
wbreza: Let's turn this into an array/map so we can support many tags on a region.

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -140,4 +140,49 @@
...
+    width: calc(100% - 40px);
+    height: 600px;
+    grid-template-columns: 64px 1fr;
+    grid-template-rows: 1fr;
+
+    margin:20px;
+    background: rgb(64,64,64);
+}
+
+#toolbarzone {
```
wbreza: FYI - long term we are going to use the toolbar integrated into the editor page vs. the canvas tools.

src/models/applicationState.ts
```diff
@@ -1,3 +1,6 @@
+import { TagsDescriptor } from "vott-ct/lib/js/CanvasTools/Core/TagsDescriptor";
```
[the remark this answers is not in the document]
PIC123: Still working to fix this. For now, since the library exports are wrapped in a single canvastools module, vscode defaults to this. Otherwise you can import canvastools module and use relative calls for types

src/models/applicationState.ts
```diff
@@ -146,8 +150,8 @@ export interface ISize {
 export interface IRegion {
     id: string;
     type: RegionType;
-    tags: ITagMetadata[];
-    points: IPoint[];
+    tags: TagsDescriptor;
```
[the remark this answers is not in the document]
PIC123: That is fair. It's not much different. I just haven't gone through standardizing the types. Will do that

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,366 @@
+import React from "react";
+import { IAssetMetadata, IRegion, RegionType, AssetState } from "../../../../models/applicationState";
+import { CanvasTools } from "vott-ct";
+import { Editor } from "vott-ct/lib/js/CanvasTools/CanvasTools.Editor";
+import { RegionData, RegionDataType } from "vott-ct/lib/js/CanvasTools/Core/RegionData";
+import { TagsDescriptor } from "vott-ct/lib/js/CanvasTools/Core/TagsDescriptor";
+import { Point2D } from "vott-ct/lib/js/CanvasTools/Core/Point2D";
+import { Tag } from "vott-ct/lib/js/CanvasTools/Core/Tag";
+
+interface ICanvasProps {
```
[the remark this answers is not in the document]
PIC123: That is true, I've been adding that for integrating with the footer

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,366 @@
...
+
+    /**
+     * @name updateTagsForSelectedRegions
+     * @description uppdates tags for currently selected region with provided tags descriptor
+     * @param {TagsDescriptor} tagsDescriptor
+     * @returns {void}
+     */
+    public updateTagsForSelectedRegions: (tagsDescriptor: TagsDescriptor) => void;
+    private editor: Editor;
+    private primaryTag: Tag;
```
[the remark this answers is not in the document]
PIC123: It definitely can and will, this was just placeholder before integration with footer. This is from the initial canvastools example

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    public componentDidMount() {
+        const ct = CanvasTools;
+        const sz = document.getElementById("editorzone") as unknown as HTMLDivElement;
+        const tz = document.getElementById("toolbarzone")as unknown as HTMLDivElement;
+
+        // @ts-ignore
+        this.editor = new ct.Editor(sz);
+        this.editor.addToolbar(tz, ct.Editor.FullToolbarSet, "./../../../images/icons/");
```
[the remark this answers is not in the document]
PIC123: Yes, for now it was placeholder until integration with footer. Will work on just removing it

src/models/applicationState.ts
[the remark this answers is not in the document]
kichinsky: I think the best options is to use some simple convertor to exclude dependency on external types. CanvasTools core classes also support init from simple (json) objects, thus we can make both type sistems compatible. E.g. you can init a Point2D from any object that has `x` and `y` numberic properties. 

I have some flexibility to update types in CanvasTools if it is required to simplify usage or conversion.

src/react/components/pages/editorPage/canvas.tsx
PIC123: This was placeholder before integrating with footer

src/react/components/pages/editorPage/editorPage.scss
PIC123: This was placeholder until integration with toolbar

PIC123: This was placeholder until integration with toolbar

src/react/components/pages/editorPage/canvas.tsx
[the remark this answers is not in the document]
kichinsky: Let me know if you need some help with reimplementing or rewiring the toolbar for the CanvasTools. The Editor class should expose all the required APIs for that, also check the actionCallback from toolbar description objects in the Editor class.

src/react/components/pages/editorPage/editorPage.scss
kichinsky: For long term:
1) can we ensure that we have something like toolbar sets, like expose tools for box selection/manipulation vs expose tools for free-form/polygon selection, etc. I think it shoulbe some how related to the current tagging job.
2) I'm thinking on some options for something like plugins for canvastools that could allow adding new custom features (e.g., new selector types like cubes). Do you think we can have some APIs in future to extend the toolbar and shortcuts registration as well?

src/react/components/pages/editorPage/editorToolbar.tsx
```diff
@@ -17,6 +18,8 @@ export interface IEditorToolbarProps {
     project: IProject;
     actions: IProjectActions;
     items: IToolbarItemRegistration[];
+    canvas: Canvas;
+    onEditorModeChange: (mode: EditorMode) => void;
```
wbreza: Recommend renaming to "onItemSelected" and keeping the editor out of the toolbar as much as possible.  Then we can link directly into the `onToolbarItemSelected` method below to auto raise this event handler vs. each button having to implement it on its own.

wbreza: I think this would then allow us to reduce the coupling and `canvas` wouldn't be required anymore.  The editorPage itself can then listen for the `onItemSelected` event to then call into the canvas to set the selection mode.

src/react/components/toolbar/toolbarItem.tsx
```diff
@@ -1,6 +1,7 @@
 import React from "react";
-import { IProject } from "../../../models/applicationState";
+import { IProject, RegionType, EditorMode } from "../../../models/applicationState";
```
wbreza: If we make some of the changes mentioned above we likely don't need any of these changes.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,416 @@
+import React from "react";
+import * as shortid from "shortid";
+import { IAssetMetadata, IRegion, RegionType,
+        AssetState, EditorMode, IProject } from "../../../../models/applicationState";
+import { CanvasTools } from "vott-ct";
+import { Editor } from "vott-ct/lib/js/CanvasTools/CanvasTools.Editor";
```
wbreza: Let's add an item to the backlog for canvas tools so we don't need to put the paths down to all these sub areas.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,416 @@
...
+import { RegionData, RegionDataType } from "vott-ct/lib/js/CanvasTools/Core/RegionData";
+import { TagsDescriptor } from "vott-ct/lib/js/CanvasTools/Core/TagsDescriptor";
+import { Point2D } from "vott-ct/lib/js/CanvasTools/Core/Point2D";
+import { Tag } from "vott-ct/lib/js/CanvasTools/Core/Tag";
+
+export interface ICanvasProps {
```
wbreza: Would recommend adding `selectedTags` here to handle integration between tags selected in the editor page vs. how the canvas reacts to that.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -0,0 +1,416 @@
...
+     * @name onRegionSelected
+     * @description Method called when deleting a region from the editor
+     * @param {string} id the id of the deleted region
+     * @param {boolean} multiselection boolean whether multiselect is active
+     * @returns {void}
+     */
+    public onRegionSelected = (id: string, multiselect: boolean) => {
+        const currentAssetMetadata = this.props.selectedAsset;
+        if (multiselect) {
+            currentAssetMetadata.selectedRegions.push(
```
wbreza: Recommend you move `selectedRegions` to be internal state to the canvas component vs messing with the assetMetadata.

src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+//     onRegionDelete: jest.fn(),
+//     onRegionSelected: jest.fn(),
+//     addContentSource: jest.fn()
+// }))
+
+describe("Editor Canvas", () => {
+    let wrapper: ReactWrapper<ICanvasProps, {}, Canvas> = null;
+    const onAssetMetadataChanged = jest.fn();
+
+    function createTestRegionData() {
```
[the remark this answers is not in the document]
PIC123: So the difference here is that RegionData is a canvastools datatype that is used for creating the region. The test region defined in the mock factory is an IRegion mock It's separated out here but can be moved

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -193,6 +230,35 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         });
     }
 
+    private  onToolbarItemSelected(toolbarItem: ToolbarItem) {
+        switch (toolbarItem.props.name) {
+            case "drawRectangle":
```
wbreza: Instead of calling the same method over and over you can set a var with the SelectionMode then call the method once at the end of the switch block.

src/react/components/pages/editorPage/editorToolbar.tsx
```diff
@@ -2,10 +2,11 @@ import React from "react";
 import _ from "lodash";
 import { IToolbarItemRegistration } from "../../../../providers/toolbar/toolbarItemFactory";
 import IProjectActions from "../../../../redux/actions/projectActions";
-import { IProject } from "../../../../models/applicationState";
+import { IProject, EditorMode } from "../../../../models/applicationState";
 import { IToolbarItemProps, ToolbarItem, ToolbarItemType } from "../../toolbar/toolbarItem";
 import "./editorToolbar.scss";
 import { Select } from "../../toolbar/select";
+import Canvas from "./canvas";
```
wbreza: Canvas is not needed here anymore

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -156,9 +162,28 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
      * Called when a tag from footer is clicked
      * @param tag Tag clicked
      */
-    public onTagClicked(tag: ITag) {
-        // Stub for now, waiting for Phil's PR
-        return;
+    public onTagClicked(tag: ITagMetadata) {
```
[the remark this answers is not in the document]
wbreza: Let's log a bug to clean this up.

src/react/components/toolbar/select.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+            group: "canvas",
+            type: ToolbarItemType.State,
+            actions,
+            active: true,
+            project: testProject,
+            onClick: clickHandler,
+        };
+    }
+
+    beforeAll(() => {
```
wbreza: Remove references to canvas within toolbar items since it is no longer needed.

PR #420, opened 2018-12-26
Added app version to the title. Fix to copy-pasting regions with new RegionData class

kichinsky:


src/index.js
```diff
@@ -20,8 +20,18 @@ var saveState,
     assetFolder,
     sourceDir; 
 
+function setAppTitle(text) {
+  ipcRenderer.send('setAppTitle', text);
+}
+
+ipcRenderer.on('setAppTitleWithVersion', (event, message) => {
```
PIC123: Are you using this title with version anywhere? It seems like it's mainly just setting the title everywhere using setAppTitle, which doesn't include the version

kichinsky: It is a two-way pingpong:
**index.js:** setAppTitle(text) -> `setAppTitle` event
**main.js:** `setAppTitle` event -> add version info -> `setAppTitleWithVersion` event
**index.js:** `setAppTitleWithVersion` event -> update title in the head.

main.js is not aware on the html content,
index.js is not aware on the package info

kichinsky:
Ignore latest commits with filters. I will move them to another PR.

PR #439, opened 2019-01-03
Unit Test Improvements

wbreza:
- Updated file system unit tests to use mock file system
- Added missing test coverage in `LocalFileSystemFileSystem.selectContainer`
- Removed unnecessary styles in tagEditorModal
- Moved tagEditorModal buttons to modal footer


src/react/components/common/tagsInput/tagEditorModal/tagEditorModal.tsx
```diff
@@ -3,21 +3,21 @@ import Form from "react-jsonschema-form";
 import { Button, Modal, ModalBody, ModalHeader } from "reactstrap";
 import { addLocValues } from "../../../../../common/strings";
 import { ITag } from "../../../../../models/applicationState";
-import "./tagEditorModal.scss";
+import ModalFooter from "reactstrap/lib/ModalFooter";
 // tslint:disable-next-line:no-var-requires
 const formSchema = addLocValues(require("./tagEditorModal.json"));
 
 const customStyles = {
```
tbarlow12: Did you mean to remove this? You mentioned in the description that you removed unnecessary styles from this modal. Looks like this one is still there

src/services/assetService.test.ts
```diff
@@ -1,46 +1,123 @@
...
-        expect(asset).not.toBeNull();
-        expect(asset.id).toEqual(expect.any(String));
-        expect(asset.name).toEqual("asset1.jpg");
-        expect(asset.type).toEqual(AssetType.Image);
-        expect(asset.path).toEqual(path);
-        expect(asset.format).toEqual("jpg");
-    });
+    describe("Static Methods", () => {
+        it("creates an asset from a file path", () => {
+            const path = "C:\\dir1\\dir2\\asset1.jpg";
```
tbarlow12: Did we want to have some of this in the `MockFactory`? I realize we don't need to use it for every piece of fake data, but it might be nice to keep it consistent so you don't have so many string literals throughout the test. Maybe something like `MockFactory.createFilePath(ext)` or something like that

src/react/components/common/tagsInput/tagEditorModal/tagEditorModal.tsx
wbreza: Weird - i thought i removed it.  Will double check.

src/services/assetService.test.ts
wbreza: Let's do a full pass on this type of stuff and create a standalone PR.

src/react/components/common/tagsInput/tagEditorModal/tagEditorModal.tsx
wbreza: Fixed import.

PR #440, opened 2019-01-03
Elhalper/duplicate project warning

elizabethhalper:
This PR prevents users from being able to create duplicate projects. In order for the error to be thrown, the projects must both have the same name, source connection, and target connection.


src/react/components/pages/projectSettings/projectSettingsPage.tsx
```diff
@@ -68,13 +68,24 @@ export default class ProjectSettingsPage extends React.Component<IProjectSetting
             ...formData,
         };
 
-        const isNew = !(!!projectToUpdate.id);
-        await this.props.actions.saveProject(projectToUpdate);
+        const projectName = formData["name"];
```
wbreza: I would recommend that we move all this logic into the `saveProject` action and/or the `projectService` that the action uses... That will keep our save logic all in one place and have it run wherever a project is saved.

It looks like there is a way to get access to the full redux state from within an action call - https://stackoverflow.com/questions/35667249/accessing-redux-state-in-an-action-creator

Let's explore this.  Then we can have the action pull the recent project list and perform the required checks.

elizabethhalper: That makes sense. I’ll work on that today

wbreza: Also - Let's add some new tests to verify this check is working as expected.

src/react/components/pages/connections/connectionsPage.test.tsx
```diff
@@ -166,11 +162,9 @@ describe("Connections Page", () => {
 
             setImmediate(() => {
                 expect(saveConnectionSpy).toBeCalledWith(connection);
-                expect(assetProvider.initialize).toBeCalled();
```
wbreza: Any reason you are removing this?  @tbarlow12 recently added this since asset providers should be initialized while saving connections.

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -61,86 +61,6 @@ describe("Editor Page Component", () => {
         expect(loadProjectSpy).not.toBeCalled();
         expect(editorPage.prop("project")).toEqual(testProject);
     });
-
-    it("Loads project assets when state changes", (done) => {
```
wbreza: These tests shouldn't be removed either... Phil is adding new tests, but these are all still valid

src/react/components/pages/export/exportPage.test.tsx
```diff
@@ -60,7 +60,7 @@ describe("Export Page", () => {
     });
 
     it("Calls save project actions on form submit", (done) => {
-        const testProject = MockFactory.createTestProject("TestProject");
+        const testProject = MockFactory.createTestProject("55");
```
wbreza: Nit: This change shouldn't be required.

src/react/components/pages/projectSettings/projectSettingsPage.tsx
```diff
@@ -67,14 +67,11 @@ export default class ProjectSettingsPage extends React.Component<IProjectSetting
         const projectToUpdate: IProject = {
             ...formData,
         };
-
-        const isNew = !(!!projectToUpdate.id);
```
wbreza: I would still like to keep this post-save logic of redirection.

src/react/components/pages/projectSettings/projectSettingsPage.tsx
```diff
@@ -67,14 +67,11 @@ export default class ProjectSettingsPage extends React.Component<IProjectSetting
...
-        await this.props.actions.saveProject(projectToUpdate);
-
-        if (isNew) {
+        try {
+            await this.props.actions.saveProject(projectToUpdate);
             this.props.history.push(`/projects/${this.props.project.id}/edit`);
-        } else {
-            this.props.history.goBack();
+        } catch (e) {
+            console.log(e);
```
wbreza: Remove console.log for now and let error bubble... We'll handle the error handling in a future story.

src/redux/actions/projectActions.ts
```diff
@@ -16,6 +18,8 @@ export default interface IProjectActions {
     loadAssetMetadata(project: IProject, asset: IAsset): Promise<IAssetMetadata>;
     saveAssetMetadata(project: IProject, assetMetadata: IAssetMetadata): Promise<IAssetMetadata>;
 }
+// export default class ProjectActions implements IProjectActions {
```
wbreza: Remove commented code.

src/redux/actions/projectActions.ts
```diff
@@ -5,6 +5,8 @@ import { ActionTypes } from "./actionTypes";
 import { AssetService } from "../../services/assetService";
 import { ExportProviderFactory } from "../../providers/export/exportProviderFactory";
 import { createPayloadAction, IPayloadAction, createAction } from "./actionCreators";
+import store from "../store/store";
+import { rejects } from "assert";
```
wbreza: Nit: Removed unused module references.

src/redux/actions/projectActions.ts
```diff
@@ -24,11 +28,35 @@ export function loadProject(project: IProject): (dispatch: Dispatch) => Promise<
...
         const projectService = new ProjectService();
-        project = await projectService.save(project);
-        dispatch(saveProjectAction(project));
+        const projectId = project["id"];
+        const projectName = project["name"];
+        const sourceConnection = project.sourceConnection.name;
+        const targetConnection = project.targetConnection.name;
+        const projectList = getState().recentProjects;
+        if (projectList && projectList.length > 0) {
+            const isUnique = (project.id === null) ||
```
wbreza: Won't this comparison of `project.id === null` bypass the duplicate check and allow a duplicate to be created?

src/redux/actions/projectActions.ts
```diff
@@ -24,11 +28,35 @@ export function loadProject(project: IProject): (dispatch: Dispatch) => Promise<
...
+    return async (dispatch: Dispatch, getState: any) => {
         const projectService = new ProjectService();
-        project = await projectService.save(project);
-        dispatch(saveProjectAction(project));
+        const projectId = project["id"];
+        const projectName = project["name"];
+        const sourceConnection = project.sourceConnection.name;
+        const targetConnection = project.targetConnection.name;
+        const projectList = getState().recentProjects;
+        if (projectList && projectList.length > 0) {
```
wbreza: Let's wrap up this duplicate check and move it into its own function to keep the core saveProject method clean and concise.  I think it actually would fit best inside the projectService

src/react/components/pages/projectSettings/projectSettingsPage.test.tsx
```diff
@@ -1,16 +1,23 @@
...
 import { BrowserRouter as Router } from "react-router-dom";
 import MockFactory from "../../../../common/mockFactory";
 import createReduxStore from "../../../../redux/store/store";
+import { IApplicationState } from "../../../../models/applicationState";
+import initialState from "../../../../redux/store/initialState";
 import ProjectSettingsPage, { IProjectSettingsPageProps } from "./projectSettingsPage";
 
 jest.mock("../../../../services/projectService");
 import ProjectService from "../../../../services/projectService";
+// jest.mock("../../../../redux/actions/projectActions");
```
wbreza: Lots of commented out test/mock stuff here - Just checking if this was intentional.

src/react/components/pages/connections/connectionsPage.test.tsx
elizabethhalper: No--I had removed it for testing. Adding back in.

src/redux/actions/projectActions.ts
elizabethhalper: Actually as I look at this, that would make the logic much easier to implement. Thanks, I'll be working on that today :)

src/redux/actions/projectActions.ts
elizabethhalper: Yes you are right, I rewrote the logic in the projectservice

src/react/components/pages/projectSettings/projectSettingsPage.tsx
wbreza: @elizabethhalper  - Still seeing this code as removed... i don't think this should be changing with this bug fix.

src/redux/actions/projectActions.ts
```diff
@@ -24,11 +24,17 @@ export function loadProject(project: IProject): (dispatch: Dispatch) => Promise<
...
 
-export function saveProject(project: IProject): (dispatch: Dispatch) => Promise<IProject> {
-    return async (dispatch: Dispatch) => {
+export function saveProject(project: IProject):
+  (dispatch: Dispatch, getState: any) => Promise<IProject> {
+    return async (dispatch: Dispatch, getState: any) => {
         const projectService = new ProjectService();
-        project = await projectService.save(project);
-        dispatch(saveProjectAction(project));
+        const projectList = getState().recentProjects;
```
wbreza: This is great, nice and clean and concise now 👍

src/services/projectService.test.ts
```diff
@@ -99,4 +100,17 @@ describe("Project Service", () => {
 
         await expect(projectSerivce.delete(testProject)).rejects.toEqual(expectedError);
     });
+
+    it("isDuplicate returns false when called with a new project", async () => {
```
wbreza: FYI - Test name is a bit off... it can still be a duplicate if it is new if the name and connections are still the same.

src/services/projectService.ts
```diff
@@ -58,6 +59,29 @@ export default class ProjectService implements IProjectService {
         });
     }
 
+    public isDuplicate(project: IProject, projectList: IProject[]): boolean {
+        const projectId = project.id;
+        const projectName = project.name;
+        const sourceConnection = project.sourceConnection.name;
+        const targetConnection = project.targetConnection.name;
+        if (projectList && projectList.length > 0) {
+            const duplicate = (projectId === undefined) &&
```
wbreza: Won't new projects bypass this check all together now?

wbreza:
```
function isDuplicate(project: IProject, projectList: IProject[]): boolean {
    const duplicateProjects = projectList.find((p) => 
      p.id !== project.id && 
      p.name === project.name && 
      JSON.stringify(p.targetConnection.providerOptions) === JSON.stringify(project.targetConnection.providerOptions)
    );

return duplicationProjects.length > 0;
}
```

wbreza: Try something like this.

1. Same name, same id, same target connection => pass (edit scenario)
2. Same name, different id same target connection => fail (dup detected, edit project)
3. Same name, no id, same target conneciton => fail (dup detected, new project)
4. Same name, different id, different target connection => pass (no dup editing project)

src/services/projectService.test.ts
elizabethhalper: I changed it to "unique"

src/services/projectService.ts
elizabethhalper: How does case 2 happen? If this is a blocker from letting them create a duplicate project? How would a project come into existence that has a difference id with the same name and target connection?

wbreza: @elizabethhalper - through an error scenario.. you originally saved it and it was fine... then on edit you try to use a name that already exists with the same target connection.  That should trigger the failure event.

elizabethhalper: @wbreza  The fix I made is close to the code above and should address both scenarios (2 and 3)

src/react/components/pages/projectSettings/projectSettingsPage.test.tsx
```diff
@@ -38,4 +42,61 @@ describe("Project settings page", () => {
...
+                    target: project.targetConnection,
+                },
+            },
+            actions: {
+                saveProject: props.actions.saveProject,
+            },
+        });
+        wrapper.find("form").simulate("submit");
+        setImmediate(async () => {
+            // expect(saveProjectSpy).toBeCalled();
```
tbarlow12: Remove

PR #447, opened 2019-01-04
Remove StorageType from IConnection and move to IStorageProvider

tbarlow12:


src/common/mockFactory.ts
```diff
@@ -325,4 +335,15 @@ export default class MockFactory {
         const num = Math.floor(Math.random() * 255);
         return num.toString(16);
     }
+
+    private static getStorageType(providerType: string): StorageType {
+        switch (providerType) {
+            case "azureBlobStorage":
+                return StorageType.cloud;
```
wbreza: Nit: Please change these to "PascalCase", not "camelCase".

src/react/components/common/cloudFilePicker/cloudFilePicker.tsx
```diff
@@ -121,8 +121,18 @@ export class CloudFilePicker extends React.Component<ICloudFilePickerProps, IClo
         />;
     }
 
+    private isCloudConnection(connection: IConnection): boolean {
```
wbreza: Nice - this can likely get even easier after I get my PR in.  We can add additional properties at time of registration so we can lookup the providers registration options without having to instantiate it.

src/common/mockFactory.ts
wbreza: Sorry if I wasn't clear in my previous comment in your other PR.

PR #448, opened 2019-01-04
Created connection provider picker to dynamically load provider list

wbreza:
- Created connection provider picker component
- Unit tests (pending)

### DIFF
--- a/src/common/mockFactory.ts
+++ b/src/common/mockFactory.ts
@@ -3,9 +3,9 @@ import {
     IAssetMetadata, IConnection, IExportFormat, IProject, ITag, StorageType,
 } from "../models/applicationState";
 import { ExportAssetState } from "../providers/export/exportProvider";
-import { IAssetProvider } from "../providers/storage/assetProvider";
+import { IAssetProvider, IAssetProviderRegistrationOptions } from "../providers/storage/assetProvider";
 import { IAzureCloudStorageOptions } from "../providers/storage/azureBlobStorage";
-import { IStorageProvider } from "../providers/storage/storageProvider";
+import { IStorageProvider, IStorageProviderRegistrationOptions } from "../providers/storage/storageProvider";
 import { IProjectSettingsPageProps } from "../react/components/pages/projectSettings/projectSettingsPage";
 import IConnectionActions from "../redux/actions/connectionActions";
 import IProjectActions, * as projectActions from "../redux/actions/projectActions";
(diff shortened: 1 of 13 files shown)


wbreza:
@mydiemho - Code coverage report embedded in PR looks awesome!! cc @pjlittle

src/common/mockFactory.ts
```diff
@@ -322,6 +364,17 @@ export default class MockFactory {
         };
     }
 
+    /**
+     * Runs function that updates the UI, and flushes call stack
+     * @param func - The function that updates the UI
+     */
+    public static flushUi(func: () => void): Promise<void> {
```
tbarlow12: Good call to move this here

src/react/components/common/connectionProviderPicker/connectionProviderPicker.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { AssetProviderFactory } from "../../../../providers/storage/assetProvider";
+
+export interface IConnectionProviderPickerProps {
+    onChange: (value: string) => void;
+    id: string;
+    value: string;
+}
+
+export default function ConnectionProviderPicker(props: IConnectionProviderPickerProps) {
+    const storageProviders = _.values(StorageProviderFactory.providers);
```
tbarlow12: Nice and clean :)

src/registerProviders.ts
```diff
@@ -7,16 +7,37 @@ import { BingImageSearch } from "./providers/storage/bingImageSearch";
...
     // Storage Providers
-    StorageProviderFactory.register("localFileSystemProxy", (options) => new LocalFileSystemProxy(options));
-    StorageProviderFactory.register("azureBlobStorage", (options) => new AzureBlobStorage(options));
+    StorageProviderFactory.register({
+        name: "localFileSystemProxy",
+        displayName: strings.connections.providers.local.title,
+        factory: (options) => new LocalFileSystemProxy(options),
+    });
+    StorageProviderFactory.register({
+        name: "azureBlobStorage",
```
tbarlow12: Is there a central place where we can keep these strings? It would be nice to not need to type out string literals if not necessary. Open to discussion though

wbreza: Let's discuss... Now since the provider list is dynamic there really shouldn't be any other instances of this string outside of unit tests.  Happy to put these into a constant/enum if we think we need it.

PR #451, opened 2019-01-07
Optional initialization for Asset Provider on connection creation

tbarlow12:


src/react/components/pages/connections/connectionsPage.tsx
```diff
@@ -120,6 +121,10 @@ export default class ConnectionPage extends React.Component<IConnectionPageProps
     }
 
     private onFormSubmit = async (connection: IConnection) => {
+        const assetProvider = AssetProviderFactory.createFromConnection(connection);
+        if (assetProvider.initialize) {
```
wbreza: I like the idea of an initialize, but I feel it should be pushed back down into the factory itself vs. checking here if the assetProvider contains an initialization method.

src/providers/storage/assetProvider.ts
```diff
@@ -17,14 +18,15 @@ export class AssetProviderFactory {
...
     public static create(name: string, options?: any): IAssetProvider {
         Guard.emtpy(name);
 
         const handler = AssetProviderFactory.handlerRegistry[name];
         console.log(handler);
-        // if (!handler) {
-        //     throw new Error(`No asset provider has been registered with name '${name}'`);
-        // }
 
         return handler(options);
```
wbreza: I think this would be a good place to instantiate the provicer, then check if it has an initialize method... if so call it before returning the handler.  Only additional though is that it is defined as async (returns promise) which would require any callers to await the creation of the provider.  Let's discuss.

src/providers/storage/assetProvider.ts
```diff
@@ -17,14 +18,15 @@ export class AssetProviderFactory {
...
 
+    public static createFromConnection(connection: IConnection): IAssetProvider {
+        return this.create(connection.providerType, connection.providerOptions);
+    }
+
     public static create(name: string, options?: any): IAssetProvider {
         Guard.emtpy(name);
 
         const handler = AssetProviderFactory.handlerRegistry[name];
         console.log(handler);
```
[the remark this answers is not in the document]
tbarlow12: Will do

src/react/components/pages/connections/connectionsPage.tsx
tbarlow12: Sounds good. Just in the create?

src/providers/storage/assetProvider.ts
tbarlow12: Flag in create to initialize or not? But I guess the method signature would still need to be async

src/providers/storage/azureBlobStorage.test.ts
```diff
@@ -187,4 +187,35 @@ describe("Azure blob functions", () => {
...
+        expect(serviceURL.prototype.listContainersSegment).toBeCalled();
+    });
+
+    it("throws an error if container not found and not created", async () => {
+        const newContainerName = "newContainer";
+        const provider: AzureBlobStorage = new AzureBlobStorage({
+            ...options,
+            containerName: newContainerName,
+        });
+        expect(provider.initialize()).rejects.toContain({
```
wbreza: Missing await here?

src/react/components/pages/connections/connectionsPage.tsx
wbreza: Thought we were removing this from here since it's in the save method now?

src/providers/storage/azureBlobStorage.test.ts
tbarlow12: This is just how jest handles testing rejected promises... But I'm going to just add a try catch to make it a little more normal, this current method of testing still logs an unhandled catch error, even though the expect is supposed to catch the reject. Lots of complaints online about it

PR #457, opened 2019-01-09
production app homepage rendering bug fix AB#16462

luisamiranda:
fixed error in react-router-dom and electron interaction that prevented initial homepage loading


src/react/components/shell/mainContentRouter.tsx
```diff
@@ -12,6 +12,7 @@ import ProfileSettingsPage from "../pages/profileSettingsPage";
 export default function MainContentRouter() {
     return (
         <div className="app-content">
+            {window.location.pathname.includes('index.html') && <Redirect to="/" />}
```
wbreza: Instead of hard coding in `index.html` can we introduce a catch all route that redirects to the homepage?

wbreza: I'd also like to introduce some unit tests to verify this doesn't sneak back up on us at some point.

wbreza: Looks like you also have a minor ts linting issue.

src/react/components/shell/mainContentRouter.tsx
```diff
@@ -12,16 +12,19 @@ import ProfileSettingsPage from "../pages/profileSettingsPage";
...
+                <Route path="/settings" component={AppSettingsPage} />
+                <Route path="/profile" component={ProfileSettingsPage} />
+                <Route path="/connections/:connectionId" component={ConnectionPage} />
+                <Route path="/connections" exact component={ConnectionPage} />
+                <Route path="/projects/:projectId/edit" component={EditorPage} />
+                <Route path="/projects/create" component={ProjectSettingsPage} />
+                <Route path="/projects/:projectId/settings" component={ProjectSettingsPage} />
+                <Route path="/projects/:projectId/export" component={ExportPage} />
+                <Route path="/projects/:projectId/active-learning" component={ActiveLearningPage} />
+                <Route component={HomePage} />
```
wbreza: do we need a `path="*"` attribute?  Referenced from https://stackoverflow.com/questions/32128978/react-router-no-not-found-route

src/react/components/shell/mainContentRouter.tsx
luisamiranda: fixed, fixed, and fixed.

src/react/components/shell/mainContentRouter.tsx
luisamiranda: instead of "*", I wrapped the router in a Switch, which allows fallthrough to the final, default route.

wbreza:
@luisamiranda Based on the commit history it appears you are doing "merges" into your branch from v2 vs. rebases... Let me know if you need help with the rebase process.

luisamiranda:
Recreated pull request without everyone else's changes attached. Abandoning this one.

PR #459, opened 2019-01-09
HostProcessType Provider Filters

JordanEllis6809:


src/common/hostProcess.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface IHostProcess {
+    type: HostProcessType;
+    release: string;
+}
+
+/**
+ * @enum ELECTRON - Electron Host Process Type
+ * @enum BROWSER - Browser Host Process Type
+ */
+export enum HostProcessType {
```
wbreza: I recommend we use simplier values to make this more clear and use power of 2 example from https://patrickdesjardins.com/blog/how-to-set-and-read-bitwise-enum-values-in-typescript

src/common/hostProcess.ts
```diff
@@ -0,0 +1,47 @@
...
+export function typeToFriendlyName(type: HostProcessType): string {
+    switch (type) {
+        case HostProcessType.Electron:
+            return "electron";
+        case HostProcessType.Browser:
+            return "browser";
+    }
+}
+
+import os from "os";
```
wbreza: This is not browser safe - this will likely fail in the web scenario, no?

tslint.json
```diff
@@ -9,7 +9,8 @@
         "no-console": false,
         "no-shadowed-variable": false,
         "ordered-imports": false,
-        "no-string-literal": false
+        "no-string-literal": false,
+        "no-bitwise": false
```
wbreza: When we change the bitwise enum stuff above we may be able to remove this... i don't know if this flag limits all bitwise operators or just a subset.

src/common/hostProcess.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @enum ELECTRON - Electron Host Process Type
+ * @enum BROWSER - Browser Host Process Type
+ */
+export enum HostProcessType {
+    Electron = 1 << 0,
+    Browser = 1 << 1,
+    All = ~(~0 << 2),
+}
+
+export function typeToFriendlyName(type: HostProcessType): string {
```
wbreza: Doesn't seem like its used anywhere.

src/providers/storage/assetProviderFactory.ts
```diff
@@ -32,10 +34,15 @@ export class AssetProviderFactory {
             options = {
                 name: nameOrOptions,
                 displayName: nameOrOptions,
+                platformSupport: HostProcessType.All,
                 factory,
             };
         }
 
+        if ((options.platformSupport & HostProcess.type) === 0) {
```
wbreza: Need unit tests to validate this registration change.

src/providers/storage/storageProviderFactory.ts
```diff
@@ -43,10 +44,15 @@ export class StorageProviderFactory {
             options = {
                 name: nameOrOptions,
                 displayName: nameOrOptions,
+                platformSupport: HostProcessType.All,
                 factory,
             };
         }
 
+        if ((options.platformSupport & HostProcess.type) === 0) {
```
wbreza: Need unit tests to validate this registration change.

src/registerProviders.ts
```diff
@@ -8,34 +8,40 @@ import { LocalFileSystemProxy } from "./providers/storage/localFileSystemProxy";
...
 import registerToolbar from "./registerToolbar";
 import { strings } from "./common/strings";
+import { HostProcessType } from "./common/hostProcess";
 
 export default function registerProviders() {
     // Storage Providers
     StorageProviderFactory.register({
         name: "localFileSystemProxy",
         displayName: strings.connections.providers.local.title,
+        platformSupport: HostProcessType.All,
```
wbreza: Let's default the platform support for all within the registration method... if not defined in the options we will assume it is all unless otherwise specified.

src/providers/storage/assetProviderFactory.ts
```diff
@@ -10,6 +11,7 @@ export interface IAssetProviderRegistrationOptions {
     name: string;
     displayName: string;
     description?: string;
+    platformSupport: HostProcessType;
```
wbreza: Make optional

src/common/hostProcess.ts
JordanEllis6809: IMHO I think writing out the bit-shifting is more self explanatory than writing the decimal representations (1, 2, 3). With the decimal form, I think someone unfamiliar with what's happening could easily come in and put a new entry with an incorrect number, for example:
```
enum HostProcessType {
  Electron = 1, // 001
  Browser = 2, // 010
  NEW = 3,      // 011
  All = 4,          // 100
}
```
But in reality the correct numbering should be: 1, 2, 4, 7.

JordanEllis6809: chatted offline, and I also agree the decimal for with better comments is more understandable. The set all bits syntax is a bit cryptic.

src/common/hostProcess.ts
JordanEllis6809: This is in fact browser safe, and returns the "userAgent" string.

package.json
```diff
@@ -28,20 +28,25 @@
...
-        "eject": "react-scripts eject",
-        "electron": "webpack --config ./config/webpack.dev.js && electron .",
-        "electron:prod": "webpack -p --config ./config/webpack.prod.js",
-        "electron-start": "node src/electron/start",
-        "react-start": "react-scripts start",
-        "compile": "tsc",
-        "release": "npm run build && npm run electron:prod && electron-builder",
-        "postinstall": "electron-builder install-app-deps"
+        "postinstall": "electron-builder install-app-deps",
+        "predebug": "npm run build && npm run webpack:dev"
```
wbreza: predebug? where is this used?  I don't see a normal "debug" script.

src/electron/start.js
```diff
@@ -12,7 +12,35 @@ const tryConnection = () => client.connect({ port: port }, () => {
         console.log('starting electron');
         startedElectron = true;
         const exec = require('child_process').exec;
-        exec('npm run electron');
+        const electron = exec('npm run electron:run:dev', (error, stdout, stderr) => {
```
wbreza: Thanks for adding this - this should help out when diagnosing weird electron start issues.

src/providers/storage/assetProviderFactory.ts
```diff
@@ -36,6 +38,14 @@ export class AssetProviderFactory {
             };
         }
 
+        if (options.platformSupport == null) {
```
wbreza: This would miss an undefined `platformSupport`.  Change to `if (!options.platformSupport)` which will catch both null and undefined values.

src/providers/storage/storageProviderFactory.ts
```diff
@@ -47,6 +48,14 @@ export class StorageProviderFactory {
             };
         }
 
+        if (options.platformSupport == null) {
```
wbreza: Same as above.

src/registerProviders.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        it("Doesn't Register localFileSystemProxy", () => {
+            process.env.TEST = "false";
+
+            const mock = jest.mock("os", () => {
+                return {
+                    release: () => "browser",
+                };
+            });
+
+            const registerProviders = require("./registerProviders").default;
```
wbreza: We have standardized on using `import` vs. `require`.  Please update and move module imports to top of file.

src/registerProviders.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    describe("Electron Registration", () => {
+        it("Does Register localFileSystemProxy", () => {
+            const mock = jest.mock("os", () => {
+                return {
+                    release: () => "electron",
+                };
+            });
+
+            const registerProviders = require("./registerProviders").default;
```
wbreza: Convert to use imports.

src/registerProviders.test.ts
```diff
@@ -0,0 +1,29 @@
+import registerProviders from "./registerProviders";
+import { StorageProviderFactory } from "./providers/storage/storageProviderFactory";
+import { AssetProviderFactory } from "./providers/storage/assetProviderFactory";
+import hostProcess from "./common/hostProcess";
+
+describe("Register Providers", () => {
+
+    describe("Browser Registration", () => {
+        it("Doesn't Register localFileSystemProxy", () => {
+            hostProcess.type = 2;
```
wbreza: Nice and easy 👍

src/common/hostProcess.ts
```diff
@@ -0,0 +1,38 @@
...
+export enum HostProcessType {
+    Electron = 1, // bits: 01
+    Browser = 2,  // bits: 10
+    All = 3,      // bits: 11
+}
+
+import os from "os";
+const osRelease = os.release().toLowerCase();
+
+function getHostProcessType(): HostProcessType {
```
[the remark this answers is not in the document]
tbarlow12: @JordanEllis6809 was this resolved?

src/electron/start.js
tbarlow12: This is going to be useful :)

src/common/hostProcess.ts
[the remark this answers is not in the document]
JordanEllis6809: No there was a backlog item added for it though.

Testing this requires stubbing out the os module twice within one test, which requires using `require` instead of `import`. This was an approach that was previously taken in this PR's commit history, and was something we didn't want to do as it went against our style guide. A good solution that tests what we want to test isn't immediately obvious...

PR #462, opened 2019-01-10
Tmiller load export providers

tmiller-msft:
- Updated register provider for exports

### DIFF
--- a/src/providers/export/exportProvider.test.ts
+++ b/src/providers/export/exportProvider.test.ts
@@ -10,7 +10,11 @@ describe("Export Provider Base", () => {
     it("initializes the asset and storage providers", () => {
         registerProviders();
 
-        ExportProviderFactory.register("test", (project) => new TestExportProvider(project));
+        ExportProviderFactory.register({
+            name: "test",
+            displayName: "Test DisplayNameq",
+            factory: (project) => new TestExportProvider(project),
+        });
         const exportProvider = ExportProviderFactory.create("test", testProject) as TestExportProvider;
         const assetProvider = exportProvider.getAssetProvider();
         const storageProvider = exportProvider.getStorageProvider();
--- a/src/providers/export/exportProviderFactory.ts
+++ b/src/providers/export/exportProviderFactory.ts
@@ -2,6 +2,13 @@ import Guard from "../../common/guard";
 import { IExportProvider } from "./exportProvider";
 import { IProject } from "../../models/applicationState";
 
+export interface IExportProviderRegistrationOptions {
+    name: string;
+    displayName: string;
+    description?: string;
+    factory: (project, IProject, options?: any) => IExportProvider;
+}
+
 /**
  * @name - Export Provider Factory
  * @description - Creates instance of export providers based on request providery type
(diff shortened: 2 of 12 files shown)


src/providers/export/exportProvider.test.ts
```diff
@@ -10,7 +10,11 @@ describe("Export Provider Base", () => {
     it("initializes the asset and storage providers", () => {
         registerProviders();
 
-        ExportProviderFactory.register("test", (project) => new TestExportProvider(project));
+        ExportProviderFactory.register({
+            name: "test",
+            displayName: "Test DisplayNameq",
```
wbreza: Nit: typo with "q" at the end.

src/react/components/common/exportProviderPicker/exportProviderPicker.tsx
```diff
@@ -0,0 +1,37 @@
...
+    id: string;
+    value: string;
+}
+
+export default function ExportProviderPicker(props: IExportProviderPickerProps) {
+    const exportProviders = _.values(ExportProviderFactory.handlers);
+
+    const allProviders = _([])
+        .concat(exportProviders)
+        .uniqBy("name")
```
wbreza: This unique expression is likely not required in here since we're not merging the providers with any other array.  In the storage/asset provider implementation there can be duplicate registrations because a "storage" provider is a "asset" provider.

src/react/components/pages/export/exportForm.test.tsx
```diff
@@ -3,20 +3,37 @@ import ExportForm, { IExportFormProps, IExportFormState } from "./exportForm";
...
+        });
+    });
+
     const onSubmitHandler = jest.fn();
 
     it("State is initialized without export settings", () => {
         const defaultExportType = "vottJson";
         const props: IExportFormProps = {
-            settings: null,
+            settings: {
```
wbreza: FYI - this test specifically is testing the scenario when the settings have been been previously saved, so they should be null/undefined.

src/providers/export/exportProviderFactory.ts
```diff
@@ -13,14 +20,15 @@ export class ExportProviderFactory {
```
wbreza: I'd recommend we rename `handlers` to `providers` to be consistent with the other factories that follow a similar model.

src/react/components/pages/export/exportForm.test.tsx
```diff
@@ -73,7 +95,7 @@ describe("Export Form Component", () => {
         };
 
         const props: IExportFormProps = {
-            settings: null,
+            settings: defaultExportSettings,
```
wbreza: FYI - the settings is not "required" on the form.  I would verify or add additional tests that this is still the case.

src/providers/export/exportProvider.test.ts
tmiller-msft: I almost changed this to Test DisplayName without Q, but alas, just removed the q.

src/react/components/pages/export/exportForm.test.tsx
tmiller-msft: Will update this with my final fix tests PR coming later today

src/react/components/pages/export/exportForm.test.tsx
tmiller-msft: Ditto other test comment

PR #465, opened 2019-01-11
16353 Extend TF Pascal VOC exporter

JacopoMangiavacchi:
Align v2 TensorFlow Pascal VOC exporter to all features currently supported on V1


JacopoMangiavacchi:
@wbreza test are failing again on expect(exportProjectSpy).not.toBeCalled();  I'll better investigate on this and also I'll probably need to implement a couple of more tests.

src/providers/export/tensorFlowPascalVOC.test.ts
```diff
@@ -27,9 +38,12 @@ describe("TFPascalVOC Json Export Provider", () => {
 
     const tagLengthInPbtxt = 37;
 
+    // jest.setTimeout(20000);
```
wbreza: Nit: Remove comments.

src/providers/export/tensorFlowPascalVOC.ts
```diff
@@ -64,71 +85,92 @@ export class TFPascalVOCJsonExportProvider extends ExportProvider<ITFPascalVOCJs
...
+        });
+
+        try {
+            await Promise.all(allImageExports);
+        } catch (err) {
+            console.log(err);
+        }
+    }
+
+    private async exportSingleImage(jpegImagesFolderName: string, element: IAssetMetadata): Promise<void> {
```
wbreza: This method is pretty large - I recommend you break this down into smaller easier to digest functions.

src/providers/export/tensorFlowPascalVOC.ts
```diff
@@ -155,16 +197,147 @@ item {
         const annotationsFolderName = `${exportFolderName}/Annotations`;
         await this.storageProvider.createContainer(annotationsFolderName);
 
+        const annotationTemplate = `
+<annotation verified="yes">
```
wbreza: Please move out to external file and see if we can import it like we do with JSON files.

src/providers/export/tensorFlowPascalVOC.ts
```diff
@@ -155,16 +197,147 @@ item {
...
         // Save Annotations
-        // TODO
+        this.imagesInfo.forEach((imageInfo, imageName) => {
+            allAnnotationExports.push(
+                new Promise(async (resolve, reject) => {
+                    const imageFilePath = `${annotationsFolderName}/${imageName}`;
+                    const assetFilePath = `${imageFilePath.substr(0, imageFilePath.lastIndexOf("."))
+                        || imageFilePath}.xml`;
+
+                    const objectsXML = imageInfo.objects.map((o) => {
```
wbreza: If you convert the `%variable% notation to use javascript `${variable}` notation you can then use the same string interpolation function that we use in localization.

src/providers/export/tensorFlowPascalVOC.ts
```diff
@@ -155,16 +197,147 @@ item {
...
+
+        try {
+            await Promise.all(allAnnotationExports);
+        } catch (err) {
+            console.log(err);
+        }
     }
 
-    private async exportImageSets(exportFolderName: string, allAssets: IAssetMetadata[]) {
+    private async exportImageSets(exportFolderName: string, allAssets: IAssetMetadata[],
```
wbreza: Split out into smaller, easier to understand methods.

src/providers/export/tensorFlowPascalVOC.ts
```diff
@@ -155,16 +197,147 @@ item {
...
+
+                } else {
+                    const imageSetFileName = `${imageSetsMainFolderName}/${tag.name}.txt`;
+                    await this.storageProvider.writeText(imageSetFileName, tagsDict.get(tag.name).join("\n"));
+                }
+            });
+        }
+    }
+
+    private shuffle(a: any[]) {
```
wbreza: What does this do and how is it used?

src/react/components/pages/export/exportPage.tsx
```diff
@@ -74,6 +74,7 @@ export default class ExportPage extends React.Component<IExportPageProps> {
         };
 
         await this.props.actions.saveProject(projectToUpdate);
+        await this.props.actions.exportProject(projectToUpdate);
```
wbreza: FYI - We don't want this here right now.  It is also causing your CI test failure.  You can run export from the editor page by clicking the "Export" button in the toolbar.

src/providers/export/tensorFlowPascalVOC.ts
JacopoMangiavacchi: shuffle() it's a utility function to shuffle a generic array.  I call this method to shuffle each tag arrays before splitting in test and validation set.  In ML if you're not working with sequenced datapoint like for example time series it's a good practice to shuffle an input dataset before splitting it in test and training dataset.

Anyway I'll put some comment in the code

PR #468, opened 2019-01-11
Re-add onTagClicked

tbarlow12:
- onTagClicked was accidentally removed in refactoring process, re-adding with stub handler


JordanEllis6809:
LGTM!

src/react/components/pages/editorPage/editorFooter.tsx
```diff
@@ -6,6 +6,7 @@ import { debug } from "util";
 export interface IEditorFooterProps {
     tags: ITag[];
     onTagsChanged: (value) => void;
+    onTagClicked: (value) => void;
```
wbreza: Is is required or optional?  I see you re-introduced the props in the test but didn't assert it was called.  I think most of the event handle style props should be marked as optional even though we likely would always want to handle it specifically within VoTT.

wbreza: We should have a test to validate it is called as expected.

PR #470, opened 2019-01-11
Azure Custom Vision Export Provider - Phase 2

wbreza:
- Completed implementation
- New Unit tests for export scenarios


src/common/htmlFileReader.ts
```diff
@@ -38,6 +39,26 @@ export default class HtmlFileReader {
         }
     }
 
+    /**
+     * Downloads the binary blob from the blob path
+     * @param asset The asset to download
+     */
+    public static async getAssetBlob(asset: IAsset): Promise<Blob> {
```
wbreza: Need tests.

src/common/utils.ts
```diff
@@ -26,3 +26,18 @@ export const KeyCodes = {
     shift: 16,
     tab: 9,
 };
+
+/**
+ * Generates a query string from the key/values of a JSON object
+ * @param object The json object
+ * @returns A value representing a URL compatible query string
+ */
+export function createQueryString(object: any): string {
```
wbreza: Need tests

src/models/applicationState.ts
```diff
@@ -158,7 +158,8 @@ export interface IRegion {
     id: string;
     type: RegionType;
     tags: ITagMetadata[];
-    points: IPoint[];
+    points?: IPoint[];
```
wbreza: @PIC123 - Let's discuss how we handle the different region types.

src/react/components/toolbar/exportProject.tsx
```diff
@@ -1,7 +1,17 @@
 import { ToolbarItem } from "./toolbarItem";
+import { toast } from "react-toastify";
 
 export class ExportProject extends ToolbarItem {
-    protected onItemClick() {
-        this.props.actions.exportProject(this.props.project);
+    protected async onItemClick() {
```
wbreza: Need tests.

src/common/htmlFileReader.ts
tbarlow12: @wbreza ^ are tests coming in this PR or coming PR?

src/common/mockFactory.ts
```diff
@@ -417,6 +423,44 @@ export default class MockFactory {
         };
     }
 
+    public static createAzureCustomVisionTags(count: number = 10): IAzureCustomVisionTag[] {
```
tbarlow12: Nit: jsdocs?

src/providers/export/azureCustomVision.test.ts
```diff
@@ -1,26 +1,42 @@
...
-        testProject = MockFactory.createTestProject("TestProject");
-        testProject.exportFormat = {
-            providerType: "azureCustomVision",
-            providerOptions: {},
-        };
-
-        axios.post = jest.fn(() => Promise.resolve({
-            data: {
-                id: expect.any(String),
+        jest.resetAllMocks();
```
tbarlow12: This looks useful. Hadn't seen it before

src/common/htmlFileReader.ts
wbreza: this one - just notes to myself :)

src/providers/export/azureCustomVision.test.ts
```diff
@@ -1,26 +1,42 @@
...
+            ...MockFactory.createTestProject("TestProject"),
+            assets: {
+                "asset-1": MockFactory.createTestAsset("1", AssetState.Tagged),
+                "asset-2": MockFactory.createTestAsset("2", AssetState.Tagged),
+                "asset-3": MockFactory.createTestAsset("3", AssetState.Visited),
+                "asset-4": MockFactory.createTestAsset("4", AssetState.NotVisited),
             },
-        }));
+            exportFormat: {
+                providerType: "azureCustomVision",
```
tbarlow12: Nit: Instantiate provider options in MockFactory

src/providers/export/azureCustomVision.test.ts
```diff
@@ -46,15 +71,15 @@ describe("Azure Custom Vision Export Provider", () => {
...
-        expect(axios.post).toBeCalledWith(
-            // tslint:disable-next-line:max-line-length
-            expect.stringContaining("https://southcentralus.api.cognitive.microsoft.com/customvision/v2.2/Training/projects?"),
-            null,
-            expect.objectContaining({
-                headers: {
-                    "Training-key": customVisionOptions.apiKey,
-                },
-            }));
+        const customVisionProject: IAzureCustomVisionProject = {
```
tbarlow12: Same as above. I don't know for sure if we'll be re-using these test objects, but it would be nice to have them in place as the project grows and more tests are written

src/providers/export/azureCustomVision.test.ts
```diff
@@ -93,6 +118,159 @@ describe("Azure Custom Vision Export Provider", () => {
...
 
         expect(newOptions).toEqual(customVisionOptions);
-        expect(axios.post).not.toBeCalled();
+        expect(AzureCustomVisionService.prototype.create).not.toBeCalled();
+    });
+
+    describe("Export Scenarios", () => {
+        beforeEach(() => {
+            AssetService.prototype.getAssetMetadata = jest.fn((asset: IAsset) => {
+                const regions: IRegion[] = [
```
tbarlow12: Same as above

src/providers/export/azureCustomVision.ts
```diff
@@ -15,62 +25,201 @@ export interface IAzureCustomVisionOptions {
...
-        const config: AxiosRequestConfig = {
-            headers: {
-                "Training-key": customVisionOptions.apiKey,
-            },
-        };
-
-        // tslint:disable-next-line:max-line-length
-        const url = `https://southcentralus.api.cognitive.microsoft.com/customvision/v2.2/Training/projects?${queryString}`;
-        const response = await axios.post(url, null, config);
+        customVisionProject = await this.customVisionService.create(customVisionProject);
```
tbarlow12: Nice :octocat:

src/providers/export/azureCustomVision/azureCustomVisionService.ts
```diff
@@ -0,0 +1,168 @@
+import axios, { AxiosRequestConfig } from "axios";
+import { createQueryString } from "../../../common/utils";
+import Guard from "../../../common/guard";
+
+export interface IAzureCustomVisionServiceOptions {
```
tbarlow12: Nit: JSDocs

src/providers/export/azureCustomVision/azureCustomVisionService.ts
```diff
@@ -0,0 +1,168 @@
...
+        Guard.emtpy(projectId);
+        Guard.null(contents);
+
+        const url = `${this.options.baseUrl}/projects/${projectId}/images`;
+        const config = this.createRequestConfig();
+        config.headers["Content-Type"] = "application/octet-stream";
+        const response = await axios.post(url, contents, config);
+
+        if (response.status !== 200) {
+            throw new Error("Error saving image");
```
tbarlow12: More of a question - do we add localization for error messages?

src/providers/export/azureCustomVision.test.ts
mydiemho: @tbarlow12, why is this not recommended?

src/providers/export/azureCustomVision.ts
```diff
@@ -15,62 +25,201 @@ export interface IAzureCustomVisionOptions {
...
+     * @returns The upload results
+     */
+    public async export(): Promise<IExportResults> {
+        const customVisionTags = await this.syncTags();
+        const assetsToExport = await this.getAssetsForExport();
+        const tagMap = _.keyBy(customVisionTags, "name");
+
+        const createImageTasks = assetsToExport.map((asset) => {
+            return this.uploadAsset(asset, tagMap)
+                .then(() => {
```
mydiemho: could we use `await` here too instead of `then`?

src/react/components/toolbar/saveProject.tsx
```diff
@@ -1,7 +1,13 @@
 import { ToolbarItem } from "./toolbarItem";
+import { toast } from "react-toastify";
```
mydiemho: what's the difference between toast and the messageBox class we have?

src/providers/export/azureCustomVision.ts
wbreza: No - this map is creating an array of tasks which are then processed in parallel below... The then/catch here is wrapping the response into a result object since some items can fail.  This allows the Promise.all() to always wait and return a resolved promise with the full results.

src/react/components/toolbar/saveProject.tsx
wbreza: A `toast` is a message that is display to the user typically in the corner of a page and goes away after a few seconds automatically.  An alert/message box is a block UI dialog that requires user interaction to dismiss.

wbreza:
@mydiemho - Not really understanding why you are requesting changes.  This review is already broken down and this is not actually a large review.  The feature is large and this is just the pieces related to export.

src/models/applicationState.ts
```diff
@@ -169,7 +170,22 @@ export interface IRegion {
...
+
+/**
+ * @name - Bouding Box
+ * @description - Defines the tag usage within a bounding box region
+ * @member left - Defines the left x boundary for the start of the bounding box
+ * @member top - Defines the top y boundary for the start of the boudning box
+ * @member width - Defines the width of the bounding box
+ * @member height - Defines the height of the bounding box
+ */
+export interface IBoundingBox {
```
JacopoMangiavacchi: It is not confusing/denormalized to have a specific boundingBox property when we already have the points array?  

We should probably just implement an utility function that return an IBoundingBox interface from the points array.  This function should work with all region types.  For polygon eventually just taking the almost left-right-up-down coordinates and create a box with these.

src/providers/export/azureCustomVision.test.ts
JacopoMangiavacchi: Not completely sure we should put anything in MockFactory.  This configuration seems specific to azure-custom-vision so probably it will not be reused by other tests.

src/providers/export/azureCustomVision.ts
```diff
@@ -15,62 +25,201 @@ export interface IAzureCustomVisionOptions {
...
+    private customVisionService: AzureCustomVisionService;
+    private assetService: AssetService;
+
+    constructor(project: IProject, options: IAzureCustomVisionExportOptions) {
         super(project, options);
         Guard.null(options);
+
+        const cusomVisionServiceOptions: IAzureCustomVisionServiceOptions = {
+            apiKey: options.apiKey,
+            baseUrl: "https://southcentralus.api.cognitive.microsoft.com/customvision/v2.2/Training",
```
JacopoMangiavacchi: We should probably let the user provides also the baseUrl.  Different Azure Regions / DC may have different url

src/providers/export/azureCustomVision.test.ts
wbreza: @JacopoMangiavacchi - There are many cases were we are reusing a lot of mock data especially for core components / concepts like projects, connections, assets, etc.  Agree there should likely be a rule where if we're not going to reuse it - then it may not be worth putting on the mock factory.

src/models/applicationState.ts
wbreza: @JacopoMangiavacchi I think this warrents a bigger discussion to ensure that our `IRegion` implementation is good enough for all current region types.  Otherwise I'm fine with making this change and is also why I mentioned in another comment that we should have a discussion about this.

src/providers/export/azureCustomVision.ts
wbreza: Agreed as this service transitions into public preview... right now it is only supported via this region base url.

src/models/applicationState.ts
wbreza: @JacopoMangiavacchi

src/providers/export/azureCustomVision.test.ts
tbarlow12: It's not a huge deal either way. We've just run into quite a few scenarios where test data specific to one test gets re-used in another area. Example: I needed to test the filtering of the connections based on the provider type. So while it wasn't specific to any specific provider, I needed instantiations of connections for each provider to validate that it was filtering properly. I can see a similar use case happening here when validating export logic

src/models/applicationState.ts
wbreza: @PIC123 @JacopoMangiavacchi Will refactor in seperate PR based on our share/points/region discussion this morning.

src/models/applicationState.ts
wbreza: Will refactor in new PR.

src/providers/export/azureCustomVision/azureCustomVisionService.ts
wbreza: Will run a localization pass with a new PR to catch all of these.

PR #471, opened 2019-01-11
Remove extra text from tag on event

tbarlow12:
- Quick patch for removing the indices (`[1]  `) displayed with tag text in the onClickEvents
- Structure of event is different in different environments, hence the various places in the data it could be located

### DIFF
--- a/src/react/components/common/tagsInput/tagsInput.tsx
+++ b/src/react/components/common/tagsInput/tagsInput.tsx
@@ -146,11 +146,19 @@ export default class TagsInput<T extends ITagsInputProps> extends React.Componen
         return match;
     }
 
+    /**
+     * Get the selected tag name from event and remove index if present
+     * @param event onClick or keyPress event
+     */
     protected getTagText(event): string {
+        let text;
         if (event.target.lastChild) {
-            return event.target.lastChild.data;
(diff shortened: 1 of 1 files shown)


src/react/components/common/tagsInput/tagsInput.tsx
```diff
@@ -146,11 +146,19 @@ export default class TagsInput<T extends ITagsInputProps> extends React.Componen
         return match;
     }
 
+    /**
+     * Get the selected tag name from event and remove index if present
+     * @param event onClick or keyPress event
+     */
     protected getTagText(event): string {
```
wbreza: Instead of parsing out the text we could also set a different attr on the element... either use the id field or set a `data-` attribute that we can lookup later.

wbreza: I'd also like to see a unit test validating this fix.

tbarlow12: Will do. The main struggle has been the inconsistency in the location of the text that I'm looking for. It's one place during tests and another when using the tool. I'll dive deeper on this and get it validated with some tests

tbarlow12:
Closing for now until we get good repro case

PR #473, opened 2019-01-13
Host Process Unit Tests

wbreza:
- Refactored host process to not run inline
- Added unit tests for host process
- Updated test scripts due to jest plugin debug breakage

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,8 +42,8 @@
         "release": "npm run build && npm run webpack:prod && electron-builder",
         "pretest": "./node_modules/.bin/tslint 'src/**/*.ts*'",
         "lintfix": "./node_modules/.bin/tslint 'src/**/*.ts*' --fix",
-        "test": "cross-env TEST=true react-scripts test --env=jsdom --silent",
-        "test:ci": "cross-env TEST=true CI=true npm run test",
+        "test": "react-scripts test --env=jsdom --silent",
+        "test:ci": "set CI=true && npm run test",
         "test:coverage": "npm run test -- --coverage",
         "postinstall": "electron-builder install-app-deps",
         "predebug": "npm run build && npm run webpack:dev"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -42,8 +42,8 @@
         "release": "npm run build && npm run webpack:prod && electron-builder",
         "pretest": "./node_modules/.bin/tslint 'src/**/*.ts*'",
         "lintfix": "./node_modules/.bin/tslint 'src/**/*.ts*' --fix",
-        "test": "cross-env TEST=true react-scripts test --env=jsdom --silent",
-        "test:ci": "cross-env TEST=true CI=true npm run test",
+        "test": "react-scripts test --env=jsdom --silent",
```
wbreza: @JordanEllis6809 - Reverted these changes - it was causing the jest plugin to break when clicking the "debug" links.

src/common/hostProcess.test.ts
```diff
@@ -0,0 +1,41 @@
+import getHostProcess, { HostProcessType } from "./hostProcess";
+
+jest.mock("os");
```
wbreza: @JordanEllis6809 - This is an example on how to mock out the os module.

src/common/hostProcess.ts
```diff
@@ -19,20 +21,19 @@ export enum HostProcessType {
     All = 3,      // bits: 11
 }
 
-import os from "os";
```
wbreza: I believe this is what caused some of your initial test issues... the os module and call to release was running inline within the module therefore making it really difficult to mock since it was always referenced an initialized before any tests were running.  Moved this call to inside the method and exported the method vs. the type.

tmiller-msft:
lgtm

package.json
JordanEllis6809: Seriously sorry about introducing this, hopefully it was easy to find this root cause.

PR #476, opened 2019-01-15
Standardize on POSIX Slash path separator and use Path.resolve() to support Windows convertion at runtime

JacopoMangiavacchi:
**WIP, do not review yet. Need to test yet on Windows**

### DIFF
--- a/src/react/components/pages/editorPage/assetPreview.tsx
+++ b/src/react/components/pages/editorPage/assetPreview.tsx
@@ -33,11 +33,11 @@ export default class AssetPreview extends React.Component<IAssetPreviewProps, IA
                     </div>
                 }
                 {asset.type === AssetType.Image &&
-                    <img src={asset.path} onLoad={this.onAssetLoad} />
+                    <img src={`file://#${asset.path.replace(/(^\w+:|^)\/\//, '')}`} onLoad={this.onAssetLoad} />
                 }
                 {asset.type === AssetType.Video &&
                     <video onLoadedData={this.onAssetLoad}>
-                        <source src={`${asset.path}#t=5.0`} />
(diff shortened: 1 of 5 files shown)


elizabethhalper:
@JacopoMangiavacchi the CI failure is not due to our changes; there is a bug in v2 that is getting worked on that will block this PR until it is fixed

src/react/components/pages/editorPage/assetPreview.tsx
```diff
@@ -45,11 +33,11 @@ export default class AssetPreview extends React.Component<IAssetPreviewProps, IA
                     </div>
                 }
                 {asset.type === AssetType.Image &&
-                    <img src={asset.path} onLoad={this.onAssetLoad} />
+                    <img src={`file://#${asset.path.replace(/(^\w+:|^)\/\//, "")}`} onLoad={this.onAssetLoad} />
```
JacopoMangiavacchi: This is always prefixing asset path with the file:// protocol but if the sourceInput is configured with something like bingImageSearch the asset path strings should already have the http:// or https:// protocol string and so we should not replace this with file://

elizabethhalper:
Looks good for all connections! @JacopoMangiavacchi should someone with Windows check our changes or are we okay to merge?

src/react/components/pages/editorPage/assetPreview.tsx
```diff
@@ -59,6 +47,13 @@ export default class AssetPreview extends React.Component<IAssetPreviewProps, IA
         );
     }
 
+    private checkAssetPathProtocol(assetPath: string): string {
```
wbreza: Instead of having this check in the asset preview I'd recommend we push this logic back down into the underlying assetProvider to ensure the data is in a correct format before it is bound to the UI.  `IAssetProvider` has a `getAssets()` method.  I think this should be the location for this change.

PR #480, opened 2019-01-16
[AB#16434] modify messageBox and alert to accept show prop

mydiemho:


src/react/components/common/messageBox/messageBox.tsx
```diff
@@ -74,6 +75,16 @@ export default class MessageBox extends React.Component<IMessageBoxProps, IMessa
         });
     }
```
wbreza: Remove unused arguments (then you can remove your tslint disable

src/react/components/common/alert/alert.tsx
```diff
@@ -31,14 +32,19 @@ export default class Alert extends React.Component<IAlertProps, IAlertState> {
 
     public render() {
         return (
-            <MessageBox ref={this.messageBox}
+            <MessageBox
```
wbreza: Nit: White-space looks funky here... recommend reverting

PR #481, opened 2019-01-16
Added crypto api to easily encrypt/decrypt data

wbreza:
- Added generateKey to generate random keys for use as encryption secrets
- Added encrypt/decrypt methods

### DIFF
--- a/src/common/crypto.ts
+++ b/src/common/crypto.ts
@@ -1,4 +1,4 @@
-import { enc, lib, LibWordArray, AES } from "crypto-js";
+import { enc, lib, AES } from "crypto-js";
 import Guard from "./guard";
 
 /**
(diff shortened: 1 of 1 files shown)


src/common/crypto.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @param message The message to encrypt
+ * @param secret The base64 encoded secret
+ */
+export function encrypt(message: string, secret: string): string {
+    Guard.emtpy(message);
+    Guard.emtpy(secret);
+
+    try {
+        const secretBytes = enc.Base64.parse(secret);
+        const iv = lib.WordArray.random(24);
```
mydiemho: Nit: is there more informative name? what is iv?

PR #484, opened 2019-01-16
Video Settings added to project settings page AB#16627

tmiller-msft:
- Adding video settings to project page
- Adding tests for video settings
- Adding default value for video settings and validating it with test
- Mocking video settings for tests
- Storing video settings within project

### DIFF
--- a/src/react/components/pages/projectSettings/projectForm.tsx
+++ b/src/react/components/pages/projectSettings/projectForm.tsx
@@ -81,9 +81,13 @@ export default class ProjectForm extends React.Component<IProjectFormProps, IPro
         };
 
         // Set the default video settings if they are not already set
-        if (this.state.formData.videoSettings === null || this.state.formData.videoSettings === undefined) {
-            this.state.formData.videoSettings = { frameExtractionRate: 15 };
-        }
+        if (this.props.project) {
+            if (this.state.formData.videoSettings === null || this.state.formData.videoSettings === undefined) {
+                this.state.formData.videoSettings = {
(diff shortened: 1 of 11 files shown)


src/react/components/pages/projectSettings/projectForm.tsx
```diff
@@ -80,6 +80,11 @@ export default class ProjectForm extends React.Component<IProjectFormProps, IPro
             },
         };
 
+        // Set the default video settings if they are not already set
+        if (this.state.formData.videoSettings === null || this.state.formData.videoSettings === undefined) {
+            this.state.formData.videoSettings = { frameExtractionRate: 15 };
```
tmiller-msft: I am not a fan of this constant, but I wasn't really sure how folks were thinking about them. I saw the constants.ts file, but I'm not sure that's what everyone would be using (I also picked this number randomly, every other frame for 30fps videos, every fourth for 60fps).

tbarlow12: Makes sense. Might be cool to have all of our default settings in one place. Whether that's constants.ts or some other place I don't know

wbreza: Mentioned above, but I believe this should be set as the "default" in the JSON schema... Then at least it is within an area related to project default settings vs. off in a constants file somewhere.

tmiller-msft: Yeah, I thought about that too, I included minimum in there, I'm fine with that if everyone is.

PR #486, opened 2019-01-16
[AB#16670] add ErrorBoundary to top level app to catch render exception

mydiemho:
also add an example on how to use the global error redux
for caught exception.

### DIFF
--- a/src/react/components/pages/homepage/homePage.tsx
+++ b/src/react/components/pages/homepage/homePage.tsx
@@ -122,16 +122,13 @@ export default class HomePage extends React.Component<IHomepageProps> {
         let projectJson: IProject;
         try {
             projectJson = JSON.parse(project);
+            await this.loadSelectedProject(projectJson);
         } catch (error) {
             this.props.appErrorActions.showError({
                 title: strings.homePage.loadProjectError.title,
                 message: strings.homePage.loadProjectError.message,
             });
-
-            return;
         }
-
-        await this.loadSelectedProject(projectJson);
     }
 
     private onProjectFileUploadError = (e, err) => {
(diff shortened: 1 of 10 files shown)


src/react/components/pages/homepage/homePage.tsx
```diff
@@ -116,8 +121,20 @@ export default class HomePage extends React.Component<IHomepageProps> {
     }
 
     private onProjectFileUpload = (e, projectJson) => {
```
wbreza: I would like to see a test for this scenario to verify your showError is called when the user attempt to upload a invalid file.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -115,9 +118,20 @@ export default class HomePage extends React.Component<IHomepageProps> {
...
-    private onProjectFileUpload = async (e, projectJson) => {
-        const project: IProject = JSON.parse(projectJson);
-        await this.loadSelectedProject(project);
+    private onProjectFileUpload = async (e, project) => {
+        let projectJson: IProject;
+        try {
+            projectJson = JSON.parse(project);
+        } catch (error) {
+            this.props.appErrorActions.showError({
+                title: "Homepage has an error",
```
wbreza: Nit: we should start localizing our known exception messages.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -115,9 +118,20 @@ export default class HomePage extends React.Component<IHomepageProps> {
...
+                message: "File is not valid json",
+            });
+
+            return;
+        }
+
+        await this.loadSelectedProject(projectJson);
```
wbreza: You can move this `await this.loadSelectedProject(projectJson);` into the `try` block which will then avoid you having to have the return statement in your `catch` block

mydiemho: that mean we could have an error that is coming from loadSelectedProject and not just from json parsing, which is fine but the error message would have to be more generic

PR #488, opened 2019-01-17
Fix crypto random test failure

wbreza:
- Fixes issue where 1 unit test randomly fails


wbreza:
@mydiemho - Is there any knobs we can turn on the codecov report to make this pass?  Coverage on the effected files is already 96% however its failing the build because the line wasn't covered.  Maybe some setting to exclude patch failure if code coverage is already above some threshold?

PR #491, opened 2019-01-17
Homepage verify project upload test

wbreza:
- Added overload to mock factory flushUi that doesn't require func
- Added new test to verify loadProject method is successfully called after valid project file upload.

### DIFF
--- a/src/react/components/pages/homepage/homePage.tsx
+++ b/src/react/components/pages/homepage/homePage.tsx
@@ -115,9 +115,9 @@ export default class HomePage extends React.Component<IHomepageProps> {
         this.cloudFilePicker.current.open();
     }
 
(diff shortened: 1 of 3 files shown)


src/react/components/pages/homepage/homePage.tsx
```diff
@@ -115,9 +115,9 @@ export default class HomePage extends React.Component<IHomepageProps> {
         this.cloudFilePicker.current.open();
     }
 
-    private onProjectFileUpload = (e, projectJson) => {
+    private onProjectFileUpload = async (e, projectJson) => {
         const project: IProject = JSON.parse(projectJson);
-        this.loadSelectedProject(project);
+        await this.loadSelectedProject(project);
```
mydiemho: why are we changing this into an async function?

PR #494, opened 2019-01-19
add npm script for using plato

mydiemho:
1. update compile target to es6

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
         "test": "react-scripts test --env=jsdom --silent",
         "test:ci": "cross-env CI=true npm run test",
         "test:coverage": "npm run test -- --coverage",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
         "test": "react-scripts test --env=jsdom --silent",
         "test:ci": "cross-env CI=true npm run test",
         "test:coverage": "npm run test -- --coverage",
+        "plato": "./node_modules/.bin/tsc --noEmit false --outDir js-src && ./node_modules/.bin/plato -r -d plato-report js-src",
```
wbreza: You shouldn't need the prefix of `node_modules` on these.  If the package is installed locally within the project node will look up the reference in the local node_modules before looking into any global modules.

PR #500, opened 2019-01-22
move errorBoundary outside Router

mydiemho:
this broke navigation render, settings doesn't load until
refresh

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Fragment } from "react";
 import { connect } from "react-redux";
 import { BrowserRouter as Router } from "react-router-dom";
 import { ToastContainer } from "react-toastify";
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -52,24 +52,26 @@ class App extends React.Component<IAppProps> {
...
-                            <MainContentRouter />
+            <Fragment>
+                <Alert title={errorTitle}
+                    message={errorMessage}
+                    closeButtonColor="info"
+                    show={showError}
+                    onClose={this.props.actions.clearError}
+                />
+                <ErrorBoundary>
+                    <Router>
```
wbreza: Why are we introducing another `Router`? We are already using `Router` inside of `<MainContentRouter />`

tmiller-msft: I'm not sure why it's there, but it's been there for a long time, wasn't introduced as part of this change.

mydiemho: I don't see Router being use in MainContentRouter, is `<Route`> the same as `<Router>`

wbreza: You guys are right - sorry 👍

PR #502, opened 2019-01-23
[AB#16776] fix error alert to not cover up app

mydiemho:


src/common/mockFactory.ts
```diff
@@ -24,12 +24,14 @@ export default class MockFactory {
      * Creates sample IAppError
      * @param {string} title to be display in  Alert
      * @param {string} message to be display in body of Alert
+     * @param {string} errorType to specify whether this is a render error or generic exception thrown
      * @returns {IAppError}
      */
-    public static createAppError(title: string = "", message: string = ""): IAppError {
+    public static createAppError(title: string = "", message: string = "", errorType: string = "generic"): IAppError {
```
wbreza: You should type this as the enum that you have defined below.

src/models/applicationState.ts
```diff
@@ -26,6 +26,12 @@ export interface IAppError {
     title?: string;
     message: string;
     errorCode?: string;
+    errorType: string;
+}
+
+export enum IAppErrorType {
```
wbreza: Enum isn't an interface, so you can drop the "I" prefix.

src/react/components/pages/homepage/homePage.test.tsx
```diff
@@ -132,6 +132,7 @@ describe("Connection Picker Component", () => {
         const expectedAppError = {
             title: "Project Loading has an error",
             message: "File is not valid json",
+            errorType: "generic",
```
wbreza: Update these errorType references to use the enum values.

PR #505, opened 2019-01-24
Fix shape save issue after toolbar refactor

PIC123:
Fixes bug where polygon was saved as rectangle on editor page

### DIFF
--- a/src/common/mockFactory.ts
+++ b/src/common/mockFactory.ts
@@ -467,7 +467,7 @@ export default class MockFactory {
             onAssetMetadataChanged: jest.fn(),
             editorMode: EditorMode.Rectangle,
         };
-        return new Canvas({canvasProps}, {});
+        return new Canvas({ canvasProps }, {});
     }
 
     public static createTestRegion(id = null) {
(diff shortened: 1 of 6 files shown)


src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -247,24 +247,31 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         switch (toolbarItem.props.name) {
             case "drawRectangle":
                 setSelectionMode(SelectionMode.RECT);
+                this.setEditorMode(EditorMode.Rectangle);
                 break;
             case "drawPolygon":
                 setSelectionMode(SelectionMode.POLYGON);
+                this.setEditorMode(EditorMode.Polygon);
                 break;
             case "drawPolyline":
```
wbreza: We don't currently have toolbar items to draw these things?  We should either remove these or add the toolbar items.  I'm leaning towards removing these since we don't have a clear end to end for how a point or polyline can be used.

PIC123: That makes sense it was just placeholder for now to show what was there

src/common/mockFactory.ts
```diff
@@ -600,6 +600,26 @@ export default class MockFactory {
...
+     */
+    public static waitForCondition(predicate: () => boolean, interval: number = 100): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            const handle = setInterval(() => {
+                try {
+                    if (predicate()) {
+                        clearInterval(handle);
+                        resolve();
+                    }
+                } catch (e) {
```
JacopoMangiavacchi: Don't you want to call clearInterval() also when catching exceptions ?

PR #506, opened 2019-01-24
Encrypt sensitive project settings

wbreza:
- New Protected input component
- Updated api key / secret fields to use protected input
- Encrypt sensitive project data on save
- Decrypt sensitive project data on load
- Refactor effected unit tests


src/react/components/pages/editorPage/editorSideBar.tsx
```diff
@@ -78,13 +78,13 @@ export default class EditorSideBar extends React.Component<IEditorSideBarProps,
...
             selectedAsset: asset,
         }, () => {
-            this.props.onAssetSelected(asset);
             this.listRef.current.forceUpdateGrid();
         });
     }
 
     private onAssetClicked(asset: IAsset) {
         this.selectAsset(asset);
+        this.props.onAssetSelected(asset);
```
mydiemho: why are we moving this here?

src/react/components/pages/projectSettings/projectSettingsPage.tsx
```diff
@@ -67,21 +73,20 @@ export default class ProjectSettingsPage extends React.Component<IProjectSetting
...
             </div>
         );
     }
 
-    private onFormSubmit = async (formData) => {
-        const projectToUpdate: IProject = {
-            ...formData,
-        };
+    private onFormSubmit = async (project: IProject) => {
+        const isNew = !(!!project.id);
```
mydiemho: is this common practice?  this looks like code smell to me

src/providers/export/azureCustomVision.test.ts
```diff
@@ -68,7 +72,10 @@ describe("Azure Custom Vision Export Provider", () => {
         };
 
         testProject.exportFormat.providerOptions = customVisionOptions;
-        const provider = new AzureCustomVisionProvider(testProject, testProject.exportFormat.providerOptions);
+        const provider = new AzureCustomVisionProvider(
+            testProject,
+            testProject.exportFormat.providerOptions as IAzureCustomVisionExportOptions,
+        );
```
mydiemho: Nit: should extract this to a helper function, seeing repeating code in other tests.

src/providers/export/azureCustomVision.test.ts
```diff
@@ -68,7 +72,10 @@ describe("Azure Custom Vision Export Provider", () => {
         };
 
         testProject.exportFormat.providerOptions = customVisionOptions;
-        const provider = new AzureCustomVisionProvider(testProject, testProject.exportFormat.providerOptions);
+        const provider = new AzureCustomVisionProvider(
+            testProject,
+            testProject.exportFormat.providerOptions as IAzureCustomVisionExportOptions,
```
mydiemho: why pass `providerOptions` when we're already passing in the` testProject`?

src/common/mockFactory.ts
```diff
@@ -158,6 +162,12 @@ export default class MockFactory {
         };
     }
 
+    public static createLocalFileSystemOptions(): ILocalFileSystemProxyOptions {
+        return {
+            folderPath: "C:\\projects\\vott\\project",
```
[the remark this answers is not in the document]
wbreza: The path is not actually used with the file system - it's just test data to use within a unit test.

src/common/utils.test.ts
```diff
@@ -32,4 +33,22 @@ describe("Helper functions", () => {
...
+        const securityToken = MockFactory.createSecurityToken("TestProject");
+
+        it("encrypt project does not double encrypt project", () => {
+            const encryptedProject = encryptProject(testProject, securityToken);
+            const doubleEncryptedProject = encryptProject(encryptedProject, securityToken);
+
+            expect(encryptedProject).toEqual(doubleEncryptedProject);
+        });
+
+        it("decrypt project detects already decrypted project settings", () => {
```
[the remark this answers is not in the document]
wbreza: Agreed - will update the test title

src/common/utils.ts
```diff
@@ -45,3 +47,78 @@ export function createQueryString(object: any): string {
...
+    if (providerOptions.encrypted) {
+        return providerOptions as ISecureString;
+    }
+
+    return {
+        encrypted: encryptObject(providerOptions, secret),
+    };
+}
+
+function decryptProviderOptions<T = IProviderOptions>(providerOptions: IProviderOptions | ISecureString, secret): T {
```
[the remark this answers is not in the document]
wbreza: The <T> in this case allows the caller to specify the return type which could be any derived implementation of IProviderOptions vs. just the base type.  The encrypt function only ever returns the ISecureString and the types don't change.

src/models/applicationState.ts
```diff
@@ -28,6 +30,10 @@ export interface IAppError {
     errorCode?: string;
 }
 
+export interface IProviderOptions {
```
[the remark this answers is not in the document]
wbreza: Good catch - Will add.

src/providers/export/azureCustomVision.test.ts
wbreza: This is an artifact of the `IExportProvider` implementation.  It requires the project to export as well as the options required for the export provider.  In this case it is part of the project settings itself.  I'm not super happy about this either but didn't want to introduce a change here.

src/providers/export/azureCustomVision.test.ts
wbreza: Will attempt to move this code up

src/react/components/common/alert/alert.test.tsx
```diff
@@ -15,13 +15,6 @@ describe("Alert component", () => {
         return mount(<Alert {...props}></Alert>);
     }
 
-    afterEach(() => {
```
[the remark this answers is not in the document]
wbreza: This was originally an artifact of enzyme not updating the wrapper component after initial render.  We are now calling the `update()` method on the wrapper which forces the enzyme to update the state.  Previously we were also manually making DOM calls against the document.  I verified this is no longer needed it makes this more consistent with other tests within the project.

src/react/components/pages/editorPage/editorSideBar.tsx
wbreza: During testing a encountered an infinite style look in some conditions of the editor page.  The `EditorSideBar` exposes an `onAssetSelected` prop as well as a `selectedAsset` prop which the editor bind to.  During the page lifecycle, the editor loads the assets for the properties.  Sets the first asset as `selectedAsset`.  When the user clicks a new image in the sidebar that triggers `onAssetSelected` event handler to fire which was then causing the editor to set its internal selected asset state, which in turn is then rebound to the sidebar. 

The fix is to only have the sidebar fire its `onAssetSelected` event handler during a user interaction flow that changes the selected asset, not via property binding - this is more consistent with how event handler work in general within web apps.

src/react/components/pages/projectSettings/projectSettingsPage.tsx
wbreza: The `!!` operator in javascript will evaluate any javascript expression and return a truthy or falsey value.  If a project is new, then the `id` property will not be set at this stage - it is handled in the project service.  

`!!project.id` will evaluate to `false` on a new project, therefore `isNew === false`.  We then use this value below to handle which page the user is transitioned to next.

I agree all the exclamation points can make this more difficult to expand... I'd be fine with inverting the condition to `const isExisting = !!project.id`.  Then it may be easier to understand.

src/common/mockFactory.ts
```diff
@@ -545,20 +558,50 @@ export default class MockFactory {
...
+                MockFactory.createSecurityToken("TestProject"),
+            ],
+        };
+    }
+
+    /**
+     * Creates a security token used for testing
+     * @param nameSuffix The name suffix to apply to the security token name
+     */
+    public static createSecurityToken(nameSuffix: string): ISecurityToken {
```
tbarlow12: Maybe a default name suffix here?

src/providers/export/azureCustomVision.test.ts
```diff
@@ -176,11 +179,10 @@ describe("Azure Custom Vision Export Provider", () => {
         });
 
         it("Uploads binaries, regions & tags for all assets", async () => {
-            testProject.exportFormat.providerOptions.assetState = ExportAssetState.All;
+            (testProject.exportFormat.providerOptions as IExportProviderOptions).assetState = ExportAssetState.All;
             const allAssets = _.values(testProject.assets);
```
tbarlow12: What does the `_` do here?

src/services/projectService.ts
```diff
@@ -35,17 +49,15 @@ export default class ProjectService implements IProjectService {
...
-
+                const storageProvider = StorageProviderFactory.createFromConnection(project.targetConnection);
                 await this.saveExportSettings(project);
+                project = encryptProject(project, securityToken);
                 await this.saveProjectFile(project);
 
                 await storageProvider.writeText(
                     `${project.name}${constants.projectFileExtension}`,
-                    JSON.stringify(project, null, 4));
+                    JSON.stringify(project, null, 4),
```
tbarlow12: I know this wasn't directly from this PR, but what's 4 here? Any way we can name that variable?

src/providers/export/azureCustomVision.test.ts
wbreza: This is lodash. https://lodash.com/

src/services/projectService.ts
wbreza: default indenting when generating the JSON string.  Allows the project file to be written with indents.

PR #512, opened 2019-01-26
Moved asset path check in AssetService

JacopoMangiavacchi:


src/react/components/pages/editorPage/assetPreview.tsx
```diff
@@ -4,27 +4,15 @@ import { strings } from "../../../../common/strings";
 import { Player, ControlBar, CurrentTimeDisplay, TimeDivider,
     BigPlayButton, PlaybackRateMenuButton, VolumeMenuButton } from "video-react";
 
-/**
```
tmiller-msft: Any particular reason we're removing the docs here?

JacopoMangiavacchi: Absolutely not and indeed I've no idea why that happened.  Re-inserted.

src/services/assetService.ts
```diff
@@ -102,7 +102,15 @@ export class AssetService {
      * Get assets from provider
      */
     public async getAssets(): Promise<IAsset[]> {
-        return await this.assetProvider.getAssets();
+        const assets = await this.assetProvider.getAssets();
```
wbreza: I would like to see a new unit test to validate this behavior before you check-in

PR #514, opened 2019-01-28
Separate TFRecords code on different file

JacopoMangiavacchi:
WIP: Not ready yet for approving

### DIFF
--- a/src/providers/export/tensorFlowRecords.ts
+++ b/src/providers/export/tensorFlowRecords.ts
@@ -79,40 +79,6 @@ export class TFRecordsJsonExportProvider extends ExportProvider<ITFRecordsJsonEx
         await this.exportRecords(exportFolderName, allAssets);
     }
 
-    private async writeTFRecord(fileNamePath: string, features: Features) {
-        try {
-            // Get Protocol Buffer TFRecords object with exported image features
-            const imageMessage = new TFRecordsImageMessage();
-            imageMessage.setContext(features);
-
-            // Serialize Protocol Buffer in a buffer
-            const bytes = imageMessage.serializeBinary();
-            const bufferData = new Buffer(bytes);
-            const length = bufferData.length;
-
-            // Get TFRecords CRCs for TFRecords Header and Footer
-            const bufferLength = getInt64Buffer(length);
-            const bufferLengthMaskedCRC = getInt32Buffer(maskCrc(crc32c(bufferLength)));
-            const bufferDataMaskedCRC = getInt32Buffer(maskCrc(crc32c(bufferData)));
-
-            // Concatenate all TFRecords Header, Data and Footer buffer
-            const outBuffer = Buffer.concat([bufferLength,
-                                             bufferLengthMaskedCRC,
-                                             bufferData,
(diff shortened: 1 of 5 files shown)


src/providers/export/tensorFlowRecords/tensorFlowBuilder.ts
```diff
@@ -0,0 +1,148 @@
...
+        feature.setBytesList(byteList);
+
+        const featuresMap = this.features.getFeatureMap();
+        featuresMap.set(key, feature);
+    }
+
+    /**
+     * @description - Return a Buffer representation of a single TFRecord
+     */
+    public releaseTFRecord(): Buffer {
```
wbreza: `releaseTFRecord` is a bit confusing as a name.  What about we rename this to `build()` ?

src/providers/export/tensorFlowRecords/tensorFlowBuilder.ts
```diff
@@ -0,0 +1,148 @@
...
+        }));
+    }
+
+    private features: Features;
+
+    constructor() {
+        this.features = new Features();
+    }
+
+    /**
```
wbreza: The signatures of these methods are very similar.  What do you think of creating an enum or similar `FeatureType` with the different features and using a single method with signature like 

`addFeature(featureType: FeatureType, key: string, value: string | number | Uint8Array)` ?

src/providers/export/tensorFlowRecords/tensorFlowBuilder.ts
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * @name - TFRecords Builder Class
+ * @description - Create a TFRecords object
+ */
+export class TFRecordsBuilder {
+    /**
+     * @records - An Array of TFRecord Buffer created with releaseTFRecord()
+     * @description - Return a Buffer representation of a TFRecords object
+     */
+    public static releaseTFRecords(records: Buffer[]): Buffer {
```
wbreza: Same here - the name is a bit confusing.  Maybe rename to some like `joinBuffers` or joinRecords` ?

src/providers/export/tensorFlowRecords/tensorFlowHelpers.ts
```diff
@@ -109,3 +109,16 @@ export function getInt32Buffer(value: number): Buffer {
 
     return new Buffer(intArray);
 }
+
+/**
+ * @s - Input string
+ * @description - Get a Uint8Array representation of an input string value
+ */
+export function  textEncode(str: string): Uint8Array {
```
wbreza: What happens with string values that are outside of UTF8 range?

src/providers/export/tensorFlowRecords.ts
```diff
@@ -256,6 +155,14 @@ export class TFRecordsJsonExportProvider extends ExportProvider<ITFRecordsJsonEx
         });
     }
 
+    private async writeTFRecords(fileNamePath: string, buffers: Buffer[]) {
```
wbreza: Looks like you set yourself up well to support the feature where we can combine output into a single tfrecords file.

PR #515, opened 2019-01-29
Pic123/optimize on region move

PIC123:
Switch to onREgionMoveEnd to optimize project actions/performance after updating canvastools package


wbreza:
@PIC123 - this is now out of date - Please refactor with `v2-video-support` branch.

PR #516, opened 2019-01-29
Updating Canvas to handle Videos and allow tagging - AB#16580

tmiller-msft:
- Hook video player into canvas
- Update styles base on pause state to allow tagging of individual frames of video
- Remove video player component from AssetPreview (more simplified)

### DIFF
--- a/src/react/components/pages/editorPage/canvas.tsx
+++ b/src/react/components/pages/editorPage/canvas.tsx
@@ -249,8 +249,6 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState> 
         }
     }
     public render() {
-        // const { loaded } = this.state;
-        // const { svgHost } = this.props;
         const { selectedAsset } = this.props;
 
         return (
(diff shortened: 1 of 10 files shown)


src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -351,38 +376,87 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
     private updateEditor = () => {
         this.deleteAllRegions();
+        // We need to check if we're looking for a video or image
+        if (this.props.selectedAsset.asset.type === AssetType.Image) {
+            this.loadImage();
+        } else if (this.props.selectedAsset.asset.type === AssetType.Video) {
+            this.loadVideo();
+        } else {
+            // We don't know what type of asset this is?
+            throw new Error(strings.editorPage.assetError);
```
JacopoMangiavacchi: Just wonder why we have an unknown asset type?

wbreza: FYI - we have a special `AppError` that we are throwing in other areas that handle error localization in 1 place.  Do a search for AppError and create a new ErrorCode that can then participate in the localization.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -351,38 +376,87 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
-            }
+            this.updateRegions();
         });
         image.src = this.props.selectedAsset.asset.path;
     }
 
+    private updateRegions() {
+        if (this.props.selectedAsset.regions.length) {
+            this.props.selectedAsset.regions.forEach((region: IRegion) => {
+                const loadedRegionData = new RegionData(region.boundingBox.left,
```
wbreza: I know this isn't your code - but would be a good idea to make a generic transform function that can covert a `IRegion` from our app model into the `RegionData` instance that canvas tool expects.

src/react/components/pages/editorPage/canvas.tsx
tmiller-msft: Added new Canvas error code

PR #518, opened 2019-01-31
Elhalper/encoded filename

elizabethhalper:
Gets rid of bug that didn't allow users to load assets with encoded file names

### DIFF
--- a/src/redux/reducers/currentProjectReducer.ts
+++ b/src/redux/reducers/currentProjectReducer.ts
@@ -24,19 +24,8 @@ export const reducer = (state: IProject = null, action: AnyAction): IProject => 
             return { ...action.payload };
         case ActionTypes.LOAD_PROJECT_ASSETS_SUCCESS:
             if (state) {
-                // () => {
-                //     this.setState({
-                //         assets: this.state.assets.map((asset) => {
-                //             asset.name = decodeURIComponent(asset.name);
-                //             asset.path = decodeURIComponent(asset.path);
-                //         })
-                //     })
-                // }
                 const currentAssets = { ...state.assets } || {};
                 action.payload.forEach((asset) => {
-                    asset.name = decodeURIComponent(asset.name);
-                    asset.path = decodeURIComponent(asset.path);
-                    console.log(asset.path);
                     if (!currentAssets[asset.id]) {
                         currentAssets[asset.id] = asset;
                     }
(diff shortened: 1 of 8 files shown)


src/redux/reducers/currentProjectReducer.ts
```diff
@@ -2,6 +2,7 @@ import _ from "lodash";
 import { ActionTypes } from "../actions/actionTypes";
 import { IProject } from "../../models/applicationState";
 import { AnyAction } from "../actions/actionCreators";
+import { AssetService } from "../../services/assetService";
```
wbreza: I don't think this line is required since you ended up not making the change here.

src/services/assetService.ts
```diff
@@ -107,6 +107,9 @@ export class AssetService {
         return assets.map((asset) => {
             if (!asset.path.toLowerCase().startsWith("http://") && !asset.path.toLowerCase().startsWith("https://")) {
                 asset.path = "file:" + asset.path;
+                if (encodeURI(asset.name) !== asset.name) {
```
wbreza: I would update this to always just return the path as the encodedURI vs. checking.

src/services/assetService.test.ts
```diff
@@ -19,6 +19,18 @@ describe("Asset Service", () => {
             expect(asset.format).toEqual("jpg");
         });
 
+        it("creates an asset from an encoded file", () => {
```
wbreza: This is a good test - but it is not testing the change you made.  We'll need to test that calls into `getAssets()` so we can assert that the path has been encoded when retrieved by the app.

elizabethhalper: I'm on it. I may be coming to you with questions

src/services/assetService.ts
```diff
@@ -107,6 +108,7 @@ export class AssetService {
         return assets.map((asset) => {
             if (!asset.path.toLowerCase().startsWith("http://") && !asset.path.toLowerCase().startsWith("https://")) {
                 asset.path = "file:" + asset.path;
```
wbreza: Unrelated: Does the `file:` need the `//` chars so it is like `file://` ??  I think its all working right now so I guess it does not.

tmiller-msft: Yeah, file: by itself just looks weird to me, but otherwise lgtm.

PR #519, opened 2019-02-01
Updated CSS for video/canvas positioning

wbreza:
-Simplified positioning to use CSS classes vs. JS style manipulation
-Updated set canvas sort to work on seeking
-Refactor to support async methods.

### DIFF
--- a/src/react/components/pages/editorPage/editorPage.scss
+++ b/src/react/components/pages/editorPage/editorPage.scss
@@ -135,6 +135,10 @@
 #ct-zone {
     position: relative;
     flex-grow: 1;
+
+    .asset-video {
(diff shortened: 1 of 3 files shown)


src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -144,14 +148,43 @@
...
+    .video-react {
+        z-index: 1;
+    }
+}
+
+canvas {
+    background-color: #1e2024 !important;
+}
+
+.video-react {
```
wbreza: Override video controls specifically not to overlap the video element so we don't have to deal with selection/editor zone falling on top of the video controls.

PR #523, opened 2019-02-01
Keyboard Manager Component

wbreza:
Allows a single place to register keyboard events across the application.  Automatically registers and deregisters event handlers as keyboard bindings go int/out of scope.

- New `KeyboardManager` component (top level context component)
- New `KeyboardBinding` component (Allows keyboard bindings within pages / components)

### DIFF
--- a/src/react/components/common/keyboardBinding/keyboardBinding.tsx
+++ b/src/react/components/common/keyboardBinding/keyboardBinding.tsx
@@ -11,9 +11,7 @@ export class KeyboardBinding extends React.Component<IKeyboardBindingProps> {
     public context!: IKeyboardContext;
     private deregisterBinding: () => void;
 
-    constructor(props, context) {
-        super(props, context);
-
+    public componentDidMount() {
         this.deregisterBinding = this.context.keyboard.addHandler(this.props.accelerator, this.props.onKeyDown);
     }
 
--- a/src/react/components/common/keyboardBinding/keyboardBinding.tsx
+++ b/src/react/components/common/keyboardBinding/keyboardBinding.tsx
@@ -1,0 +1,29 @@
+import { KeyboardContext, IKeyboardContext } from "../keyboardManager/keyboardManager";
+import React from "react";
+
+export interface IKeyboardBindingProps {
+    accelerator: string;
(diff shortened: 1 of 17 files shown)


src/react/components/common/keyboardBinding/keyboardBinding.tsx
```diff
@@ -0,0 +1,27 @@
...
+import React from "react";
+
+export interface IKeyboardBindingProps {
+    accelerator: string;
+    onKeyDown: (evt?: KeyboardEvent) => void;
+}
+
+export class KeyboardBinding extends React.Component<IKeyboardBindingProps> {
+    public static contextType = KeyboardContext;
+    public context!: IKeyboardContext;
```
elizabethhalper: Is this supposed to be a question mark or does the bang operator do something here? If so, what is it doing?

wbreza: https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci

src/react/components/toolbar/previousAsset.tsx
```diff
@@ -0,0 +1,11 @@
+import { ToolbarItem } from "./toolbarItem";
```
pjlittle: likely more trouble than what it's worth, but looking at previous/next in terms of DRY, an AssetNav component that takes direction?

wbreza: Totally makes sense 👍

src/registerToolbar.ts
```diff
@@ -70,6 +72,24 @@ export default function registerToolbar() {
         type: ToolbarItemType.Action,
     });
 
+    ToolbarItemFactory.register(PreviousAsset, {
+        name: "navigatePreviousAsset",
+        tooltip: "Previous Asset",
```
pjlittle: good for now, but we should pipe localization in here.

pjlittle:
Good stuff, just left a couple minor remarks. Digging these components and centralized, data-driven registration.

src/registerToolbar.ts
wbreza: 👍

PR #525, opened 2019-02-02
Reved package version and fixed minor bugs

wbreza:
- Reved package to 2.0.0-preview.1
- Don't open dev tools by default
- Fixed minor guard bug in keyboard manager


pjlittle:
Ah... I remember when `String.IsNullOrEmpty` was released... :heart: :broken_heart: 

LGTM :shipit:

PR #528, opened 2019-02-04
Added support for basic filters

kichinsky:


kichinsky:
I'm checking testing issues now. I will submit an update as resolved.

PR #535, opened 2019-02-05
Update release

PIC123:
Update release for v1

### DIFF
--- a/src/public/js/video-tagging/video-tagging.html
+++ b/src/public/js/video-tagging/video-tagging.html
@@ -788,7 +788,7 @@
             this.editor.RM.redrawAllRegions();
         }
         //Clears areaSelector state when moving across frames
-        this.editor.RM.onManipulationEnd();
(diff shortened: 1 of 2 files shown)


src/public/js/video-tagging/video-tagging.html
```diff
@@ -788,7 +788,7 @@
             this.editor.RM.redrawAllRegions();
         }
         //Clears areaSelector state when moving across frames
-        this.editor.RM.onManipulationEnd();
+        this.editor.RM.redrawAllRegions();
```
tbarlow12: @PIC123 did you mean to include this on this PR?

PIC123: Hey, yah that was causing issues with loading so I just wanted to update the small thing for release. I can remove it as well

tbarlow12: That's fine with me :)

PR #546, opened 2019-02-08
TFRecords import refactoring rebased on v2-video-support

JacopoMangiavacchi:


src/react/components/common/assetPreview/tfrecordAsset.tsx
```diff
@@ -0,0 +1,84 @@
...
+    tfRecordImage64: string;
+}
+
+/**
+ * React component that displays an image from a TFRecord asset file
+ */
+export class TFRecordAsset extends React.Component<ITFRecordProps, ITFRecordState> {
+    private image: React.RefObject<HTMLImageElement> = React.createRef();
+
+    constructor(props, context) {
```
wbreza: In our new components let avoid creating a constructor unless really needed.  You can define the state of the component inline within the class like
```
public state: ITFRecordState = {
    tfRecordImage64: "",
};
```

wbreza: Then you can ommit the constructor entirely.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -300,6 +300,20 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
             selectedAsset: assetMetadata,
             assets: _.values(this.props.project.assets),
         });
+
```
wbreza: I don't believe this code should be needed... the childAssets are pull out already in the `onAssetMetadataChanged` function and set on the state.

JacopoMangiavacchi:
@wbreza can you please take a look also at the TFRecordAsset unit test.  As from our chat conversation I wasn't able to reproduce in test the full react life cycle notifications.

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: the selectedAsset state must still be set here

src/react/components/common/assetPreview/tfrecordAsset.tsx
```diff
@@ -0,0 +1,81 @@
...
+ * React component that displays an image from a TFRecord asset file
+ */
+export class TFRecordAsset extends React.Component<ITFRecordProps, ITFRecordState> {
+    public state: ITFRecordState = {
+        tfRecordImage64: "",
+    };
+
+    private image: React.RefObject<HTMLImageElement> = React.createRef();
+
+    public render() {
```
wbreza: Instead of re-implementing the logic from the `ImageAsset` you should just reuse the image asset and bind a newly constructed asset object.  You can set your base64 data uri as the path in the generated asset.

PR #547, opened 2019-02-08
[Fixes AB#16951] - support up/down & w/s key for navigating assets

mydiemho:
1. replace toolbar left/right button with up/down buttons
to match keyboard
1. add support for up/down & w/s key for moving between assets
1. support multiple accelerators per binding

### DIFF
--- a/src/react/components/common/keyboardManager/keyboardRegistrationManager.test.ts
+++ b/src/react/components/common/keyboardManager/keyboardRegistrationManager.test.ts
@@ -111,7 +111,7 @@ describe("Keyboard Registration Manager", () => {
         expect(handlers1.length).toEqual(0);
 
         handlers2 = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCode2);
-        expect(handlers1.length).toEqual(0);
+        expect(handlers2.length).toEqual(0);
     });
 
     it("get handlers for unregistered key code returns emtpy array", () => {
@@ -137,23 +137,5 @@ describe("Keyboard Registration Manager", () => {
 
         expect(handler1).toBeCalledWith(keyboardEvent);
         expect(handler2).toBeCalledWith(keyboardEvent);
-    });
-
-    it("can register handler for multiple keyCode of the same eventType", () => {
-        const keyCode1 = "Ctrl+1";
-        const keyCode2 = "Ctrl+S";
-        const keyCodes = [keyCode1, keyCode2];
-        const handler = (evt: KeyboardEvent) => null;
-
-        keyboardManager.addHandler(KeyEventType.KeyDown, keyCodes, handler);
-
-        const handlers1 = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCode1);
-        const handlers2 = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCode2);
-
-        expect(handlers1.length).toEqual(1);
-        expect(handlers2.length).toEqual(1);
-
-        expect(handlers1[0]).toBe(handler);
-        expect(handlers2[0]).toBe(handler);
     });
 });
--- a/src/react/components/common/keyboardManager/keyboardRegistrationManager.test.ts
+++ b/src/react/components/common/keyboardManager/keyboardRegistrationManager.test.ts
@@ -59,52 +59,49 @@ describe("Keyboard Registration Manager", () => {
     });
 
     it("can register multiple handlers for same key code", () => {
-        const keyCodeString = "Ctrl+H";
-        const keyCodes = [keyCodeString];
+        const keyCode = "Ctrl+H";
         const handler1 = (evt: KeyboardEvent) => null;
         const handler2 = (evt: KeyboardEvent) => null;
 
-        keyboardManager.addHandler(KeyEventType.KeyDown, keyCodes, handler1);
-        keyboardManager.addHandler(KeyEventType.KeyDown, keyCodes, handler2);
+        keyboardManager.addHandler(KeyEventType.KeyDown, [keyCode], handler1);
+        keyboardManager.addHandler(KeyEventType.KeyDown, [keyCode], handler2);
 
-        const handlers = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCodeString);
+        const handlers = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCode);
         expect(handlers.length).toEqual(2);
     });
 
     it("list of handlers cannot be mutated outside of API", () => {
-        const keyCodeString = "Ctrl+K";
-        const keyCodes = [keyCodeString];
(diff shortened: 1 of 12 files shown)


src/registerToolbar.ts
```diff
@@ -75,19 +75,19 @@ export default function registerToolbar() {
     ToolbarItemFactory.register(PreviousAsset, {
         name: "navigatePreviousAsset",
         tooltip: "Previous Asset",
-        icon: "fas fa-arrow-circle-left",
+        icon: "fas fa-arrow-circle-up",
         group: "navigation",
         type: ToolbarItemType.Action,
-        accelerator: "ArrowLeft",
+        accelerators: ["ArrowUp", "w", "W"],
```
mydiemho: According to this [doc](https://www.w3.org/TR/uievents-code/#alphanumeric-section), I could use "KeyW" so this will also work for international keyboards.  However, our keyboard binding doesn't recognize "KeyW" accelerator.  @wbreza any idea?

src/react/components/toolbar/toolbarItem.tsx
```diff
@@ -73,14 +74,20 @@ export abstract class ToolbarItem extends React.Component<IToolbarItemProps> {
...
+
         return (
             <Fragment>
-                {this.props.accelerator &&
-                    <KeyboardBinding
-                        keyEventType={KeyEventType.KeyDown}
-                        accelerator={this.props.accelerator}
-                        onKeyEvent={this.onClick}
-                    />
+                {
```
wbreza: Instead making the accelerators on the toolbar item accept an array, lets move this down into the actual `KeyboardBinding` so anyone can pass 1 or more accelerators to a keyboard binding to have them mapped automatically.

src/registerToolbar.ts
wbreza: Agreed - Let's update the `KeyboardManager` to use the key code identified in this article vs. the way we are currently doing it.  Let me know if you need help with this.

src/registerToolbar.ts
```diff
@@ -75,19 +75,19 @@ export default function registerToolbar() {
...
         type: ToolbarItemType.Action,
-        accelerator: "ArrowUp",
+        accelerators: ["ArrowUp", "w", "W"],
     });
 
     ToolbarItemFactory.register(NextAsset, {
         name: "navigateNextAsset",
         tooltip: "Next Asset",
-        icon: "fas fa-arrow-circle-right",
+        icon: "fas fa-arrow-circle-down",
```
mydiemho: the latest change on v2 change the to `ArrowDown` but didn't update the icon

src/registerToolbar.ts
mydiemho: looks like event.code is [not yet supported in React](https://github.com/facebook/react/issues/14102), let's create a ticket to update this later when it's supported

src/react/components/common/keyboardManager/keyboardRegistrationManager.ts
```diff
@@ -24,12 +24,14 @@ export class KeyboardRegistrationManager {
...
      * Registers a keyboard event handler for the specified key code
      * @param keyEventType Type of key event (keydown, keyup, keypress)
-     * @param keyCode The key code combination, ex) Ctrl+1
+     * @param keyCodes a list of key code and key code combinations, ex) Ctrl+1
      * @param handler The keyboard event handler
+     *
+     * @returns a function for deregistering the handler
      */
-    public addHandler(keyEventType: KeyEventType, keyCode: string, handler: KeyboardEventHandler): () => void {
+    public addHandler(keyEventType: KeyEventType, keyCodes: string[], handler: KeyboardEventHandler): () => void {
```
wbreza: We should add a new guard to ensure the `keyCodes.length > 0`

src/registerToolbar.ts
wbreza: We are not using the React synthetic keyboard event in our implementation - we are binding directly using standard javascript event registration which has access to the "code" property.  However - after looking @ https://caniuse.com/#search=keyboardevent.code it doesn't appear that Edge supports `code` yet... argh.

src/react/components/common/keyboardManager/keyboardRegistrationManager.test.ts
```diff
@@ -33,74 +33,85 @@ describe("Keyboard Registration Manager", () => {
...
 
         // Attempt to add more handlers
         handlers.push(handler1, handler1, handler1);
 
-        const newHandlers = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCode);
+        const newHandlers = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCodeString);
         expect(newHandlers.length).toEqual(handlerCount);
     });
 
     it("can remove keyboard event handlers", () => {
```
mydiemho: @wbreza does this not qualify as also a test for an array of keycodes? i thought about adding another test but it looks very similar to this

src/registerToolbar.ts
mydiemho: you're right event.code works.  So should we update it now even though Edge has no support or wait till Edge supports it?

src/react/components/common/keyboardManager/keyboardRegistrationManager.test.ts
```diff
@@ -33,63 +33,67 @@ describe("Keyboard Registration Manager", () => {
     });
 
     it("can register handlers for same key code and different key event types", () => {
-        const keyCode = "Ctrl+H";
+        const keyCodeString = "Ctrl+H";
+        const keyCodes = [keyCodeString];
```
PIC123: Do you need both of these? Could you just use keyCode and [keyCode]?

src/react/components/common/keyboardManager/keyboardRegistrationManager.test.ts
```diff
@@ -109,21 +113,92 @@ describe("Keyboard Registration Manager", () => {
...
+            const deregister = keyboardManager.addHandler(KeyEventType.KeyDown, keyCodes, handler);
+
+            // Get registered handlers
+            let handlers1 = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCode1);
+            expect(handlers1.length).toEqual(1);
+
+            let handlers2 = keyboardManager.getHandlers(KeyEventType.KeyDown, keyCode2);
+            expect(handlers2.length).toEqual(1);
+
+            // Invode deregister functions
```
PIC123: Nit: Typo (Invoke}

mydiemho:
@wbreza  you're right event.code works. So should we update it now even though Edge has no support or wait till Edge supports it?

src/react/components/common/keyboardManager/keyboardRegistrationManager.test.ts
mydiemho: I could use [keyCode]

PR #550, opened 2019-02-08
Added v2 complexity report scripts.

pjlittle:
- Moved from `plato` --> `es6-plato`
- Running the tool (either one...) is easier as a script, using the module/API directly
- Added shell script for managing transpilation from TS --> ES6 and running report generation easily from CI
- Should work w/ existing CI scripts; still need to do work to upload this to blob to view history over time

#AB17036

complexity

### DIFF
--- a/scripts/complexity-analysis.js
+++ b/scripts/complexity-analysis.js
@@ -1,0 +1,50 @@
+
+const plato = require('es6-plato');
+
+// parse command line args
+const optionDefinitions = [
+    { name: 'src', type: String, defaultValue: './es6-src/**' },
+    { name: 'output', type: String, defaultValue: './report' }
+]
+const commandLineArgs = require('command-line-args');
+const options = commandLineArgs(optionDefinitions)
+
+// close(ish) eslint config for ES6 + React
+let lintRules = {
+    extends: [
+        'eslint:recommended',
+        'plugin:react/recommended'
+    ],
+    plugins: [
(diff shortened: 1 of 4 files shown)


scripts/complexity-analysis.js
```diff
@@ -0,0 +1,50 @@
...
+            'jsx': true
+        }
+    },
+    rules: {
+        quotes: [2, 'double']
+    }
+};
+
+// exclude all tests, toolbar/mockFactory and localization
+let exclude = /\.test|registerToolbar|en-us|es-cl|mockFactory/;
```
JacopoMangiavacchi: I'm just curious: why are we excluding tests from this report?

pjlittle: I excluded it from the V1 report as well -- our test code isn't always indicative of our true complexity and isn't code that runs in the app.  Plus, it adds a considerable amount of overhead, in terms of sheer volume of code. 

If we want to re-introduce later, that's a possibility, or even generate a report just for test code, but for now going to leave as-is.

pjlittle:
@mydiemho - `complexity-analysis.js` just replaces what we had in `package.json` and calls the Plato API directly. I found this was easier to manage, as I could more easily dictate (and comment) what files I was skipping, as well as provide custom eslint rules.

we'll still call `npm run plato` in CI, will will invoke a script that calls this js file.

PR #551, opened 2019-02-09
Integrate vott-react NPM Package AB#16643

luisamiranda:
Tanner's, Wallace's, and my changes to incorporate vott-react packages complete at last.


src/common/strings.test.ts
```diff
@@ -93,80 +93,81 @@ describe("Localization tests", () => {
             }
         });
```
wbreza: Removed commented test.

src/react/components/pages/editorPage/editorFooter.test.tsx
```diff
@@ -68,36 +65,36 @@ describe("Footer Component", () => {
...
-        wrapper.find("div.tag")
-            .first()
-            .simulate("click", { target: { innerText: originalTags[0].name }, ctrlKey: true});
-        wrapper.find("button.btn.btn-secondary").simulate("click");
-
-        expect(onChangeHandler).not.toBeCalled();
-    });
-
-    it("clicking tag without ctrl calls onTagClickHandler ", () => {
+    // it("clicking 'ok' in modal closes and calls onChangeHandler", () => {
```
wbreza: Why are these tests commented out?  Please uncomment and ensure they are passing.

src/react/components/pages/projectSettings/projectForm.tsx
```diff
@@ -125,10 +103,61 @@ export default class ProjectForm extends React.Component<IProjectFormProps, IPro
...
+                    onChange: props.onChange,
+                };
+            }),
+            tagsInput: CustomField(TagsInput, (props) => {
+                const tagsInputProps: ITagsInputProps = {
+                    tags: props.formData,
+                    onChange: props.onChange,
+                    placeHolder: strings.tags.placeholder,
+                    onCtrlTagClick: this.onTagClick,
+                    ref: this.tagsInput,
```
wbreza: why do we need this `ref` here?

src/react/components/pages/editorPage/editorTagsInput.test.tsx
```diff
@@ -2,309 +2,37 @@ import { mount } from "enzyme";
...
+                placeHolder={props.placeHolder}
+                delimiters={props.delimiters}
+                onTagClick={props.onTagClick}
+                onShiftTagClick={props.onShiftTagClick}
+                onCtrlShiftTagClick={props.onCtrlShiftTagClick}
+            />,
         );
     }
 
-    it("tags are initialized correctly", () => {
```
tbarlow12: Most of these tests should still be in here. The modal tests can be removed because the logic for opening and closing the modal is now moved up a level (either `EditorFooter` or `ProjectForm`). But these other tests should still be valid.

src/react/components/pages/editorPage/editorTagsInput.tsx
```diff
@@ -36,38 +23,20 @@ export default class EditorTagsInput extends TagsInput<IEditorTagsInputProps> {
...
-        }
-    }
-
     private indexOfTag(id: string): number {
         let index = -1;
-        if (this.state) {
-            index = this.state.tags.findIndex((tag) => tag.id === id);
-            if (index < 0) {
-                index = this.state.tags.length + 1;
+        if (this.state) { // if stat exists
```
tbarlow12: nit: "state"

src/react/components/pages/editorPage/editorTagsInput.tsx
```diff
@@ -36,38 +23,20 @@ export default class EditorTagsInput extends TagsInput<IEditorTagsInputProps> {
...
+            index = this.props.tags.findIndex((tag) => tag.name === id); // check tags in props and return index or -1
         }
-        if (index < 0) {
+        if (index < 0) { // if neither of the above were triggered, throw error
             throw new Error(`No tag by id: ${id}`);
         }
-        index += 1;
-        return (index === 10) ? 0 : index;
+        index += 1; // whatever the index, increment by one to avoid zero-base in display
+        return (index === 10) ? 0 : index; // return index or 0, because max hotkeys are 10?
```
tbarlow12: nit: remove `?`. If you were still wondering about this, yes, the max number of hot keys are 10, and we want `0` to be the last key that shows. We could probably make that a little more clear and avoid having a magic number in the function, but I think it's fine for now

src/react/components/pages/projectSettings/projectForm.tsx
tbarlow12: @wbreza there was a reason for that, and I remember discussing this together, but it's been a little bit, so I'm a little fuzzy on the reason. I think the `ref` needs to be in the properties of the `Widget` that's created for the `Form` component. Before it wasn't necessary because we weren't creating a reference in the form, but with the change to decouple the `TagsInput` and the `TagEditorModal`, we needed to create a ref for each.

src/react/components/pages/editorPage/editorFooter.test.tsx
luisamiranda: These are edited out because they are duplicated in tests on the Tags Input component in the vott-react repo

src/react/components/pages/projectSettings/projectForm.tsx
luisamiranda: I actually don't think that we want this RefObject & ref props pattern. I would like to refactor it if possible.

src/react/components/pages/editorPage/editorFooter.test.tsx
tbarlow12: They should still be included. It's more of an integration test at this point, making sure the footer is able to interact with the tags input component and pass along the appropriate data to the next level up

src/react/components/pages/editorPage/editorTagsInput.tsx
luisamiranda: I meant to remove all of these comments. they were just to help me find where an error was coming from. BUT I was think that maybe this should be incorporated into the TagsInput component in vott-react and this component removed from here entirely. The only difference is displaying the index numbers, which seems fine to have in the project settings as well.

src/react/components/pages/editorPage/editorFooter.test.tsx
tbarlow12: Also, with the refactor, `TagsInput` no longer contains the editor modal. So any tests that are testing the actual editing of tags through the modal would need to be done here and the project form

src/react/components/pages/editorPage/editorFooter.test.tsx
```diff
@@ -2,8 +2,6 @@ import { mount, ReactWrapper } from "enzyme";
 import React from "react";
 import EditorFooter, { IEditorFooterProps, IEditorFooterState } from "./editorFooter";
 import MockFactory from "../../../../common/mockFactory";
-// tslint:disable-next-line:no-var-requires
-const TagColors = require("../../common/tagsInput/tagColors.json");
```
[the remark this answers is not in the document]
luisamiranda: added file back under src/react/components/common

src/react/components/pages/editorPage/editorFooter.test.tsx
luisamiranda: Now passing

src/react/components/pages/editorPage/editorTagsInput.tsx
luisamiranda: If one of the hotkeys is going to be 0, we could just leave the zero-indexing

src/react/components/pages/editorPage/editorTagsInput.test.tsx
luisamiranda: @tbarlow12 and @wbreza I just pushed changes that make all these tests pass. a number of them just needed displayHotKeys:true removed. I chose to leave it there commented out, because I think we should put that option back in, and it's nearly all there right now.

src/react/components/pages/projectSettings/projectForm.tsx
luisamiranda: Will address on another ticket.

src/react/components/pages/editorPage/editorTagsInput.test.tsx
```diff
@@ -2,26 +2,49 @@ import { mount } from "enzyme";
...
+            />,
         );
     }
 
     it("tags are initialized correctly", () => {
         const onChangeHandler = jest.fn();
         const wrapper = createComponent({
             tags: originalTags,
-            displayHotKeys: true,
+            // displayHotKeys: true,
```
tbarlow12: Feel free to remove the commented-out prop as well :)

luisamiranda: i thought i'd leave it to possibly add back in. I think it would be good to be able to witch on and off.

tbarlow12: I agree, we might add that back in. But I think it'll be cleaner if we just update that when we actually add the property back

luisamiranda: removed comments

src/react/components/pages/editorPage/editorTagsInput.test.tsx
```diff
@@ -117,92 +133,34 @@ describe("Tags Input Component", () => {
         expect(wrapper.find(EditorTagsInput).state().tags).toHaveLength(originalTags.length);
     });
 
-    it("ctrl clicking tag opens editor modal", () => {
```
wbreza: Why are these tests removed?  Seems like a good test to validate that the editor modal open when the take is clicked.

src/react/components/pages/editorPage/editorTagsInput.test.tsx
```diff
@@ -117,92 +133,34 @@ describe("Tags Input Component", () => {
...
-        // Shows modal
-        expect(wrapper.find(EditorTagsInput).state().showModal).toBe(true);
+        expect(onTagCtrlClickHandler).toBeCalled();
         // Does not Call onTagShiftClick
         expect(onTagShiftClickHandler).not.toBeCalled();
         // Does not call onTagClick
         expect(onTagClickHandler).not.toBeCalled();
     });
 
-    it("clicking 'ok' in modal closes and calls onChangeHandler", () => {
```
wbreza: Seems like more valuable tests to keep...

src/react/components/pages/editorPage/editorTagsInput.test.tsx
luisamiranda: those tests are removed because - as tanner confirmed previously - the modal and tags input components are no longer connected to one another. the modal does not call the function on tags input. it calls it on editorFooter.

src/react/components/pages/editorPage/editorTagsInput.test.tsx
luisamiranda: same as above. this component does not control the modal.

PR #553, opened 2019-02-11
TFRecords import refactoring

JacopoMangiavacchi:


src/react/components/common/assetPreview/assetPreview.tsx
```diff
@@ -3,24 +3,16 @@ import { IAsset, AssetType } from "../../../../models/applicationState";
 import { strings } from "../../../../common/strings";
 import { ImageAsset } from "./imageAsset";
 import { VideoAsset } from "./videoAsset";
+import { TFRecordAsset } from "./tfrecordAsset";
 
 export type ContentSource = HTMLImageElement | HTMLVideoElement;
 
-/**
```
wbreza: Please re-add these js docs.

src/react/components/common/assetPreview/imageAsset.tsx
```diff
@@ -1,15 +1,9 @@
 import React from "react";
 import { IAssetProps } from "./assetPreview";
 
-/**
```
wbreza: Re-add js docs.

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -8,41 +8,25 @@ import { IAssetProps } from "./assetPreview";
 import { IAsset, AssetType, AssetState } from "../../../../models/applicationState";
 import { AssetService } from "../../../../services/assetService";
 import { CustomVideoPlayerButton } from "../../common/videoPlayer/customVideoPlayerButton";
-import { strings } from "../../../../common/strings";
```
wbreza: Re-add js docs.

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -80,13 +64,13 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
                     <VolumeMenuButton enabled order={7.2} />
                     <CustomVideoPlayerButton order={8.1}
                         accelerator="ArrowLeft"
-                        tooltip={strings.editorPage.videoPlayer.previousTaggedFrame.tooltip}
```
wbreza: Re-add localization updates.

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -123,10 +103,6 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
         }
     }
 
-    /**
```
wbreza: More js-docs

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -311,8 +311,21 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
 
         this.setState({
             selectedAsset: assetMetadata,
-            assets: _.values(this.props.project.assets),
         });
+
+        if (!asset.parent) {
```
wbreza: As mentioned previously on other PR i don't believe we need this change since the child assets are already selected in the `onAssetMetadataChanged` event

src/common/htmlFileReader.test.ts
```diff
@@ -128,4 +129,56 @@ describe("Html File Reader", () => {
...
+            try {
+                const result = await HtmlFileReader.readAssetAttributes(imageAsset);
+            } catch (error) {
+                expect(error).toEqual(new Error("Asset not supported"));
+            }
+        });
+    });
+
+    describe("Test TFRecords", () => {
+        it("Loads attributes for an tfrecord asset", async () => {
```
mydiemho: Nit: a tfrecord

src/common/htmlFileReader.test.ts
```diff
@@ -128,4 +129,56 @@ describe("Html File Reader", () => {
             expect(axios.get).toBeCalledWith(asset.path, { responseType: "blob" });
         });
     });
+
+    describe("Test non valid asset type", () => {
```
mydiemho: do you need both describe and it, they have same text

src/providers/export/tensorFlowRecords/tensorFlowHelpers.test.ts
```diff
@@ -37,4 +37,10 @@ describe("TFRecords Helper Functions", () => {
             expect(textEncode("ABC123")).toEqual(new Uint8Array([65, 66, 67, 49, 50, 51]));
```
mydiemho: make `"ABC123"` and the array contants

src/providers/export/tensorFlowRecords/tensorFlowHelpers.ts
```diff
@@ -132,6 +132,17 @@ export function textEncode(str: string): Uint8Array {
     return result;
 }
 
+/**
+ * @arr - Input Uint8Array byte array
+ * @description - Get a UTF8 string value
+ */
+export function textDecode(arr: Uint8Array): string {
+    Guard.null(arr);
```
mydiemho: what if it's empty?

src/providers/export/tensorFlowRecords/tensorFlowReader.ts
```diff
@@ -62,4 +63,28 @@ export class TFRecordsReader {
...
     }
+
+    /**
+     * @recordPos - Record Position
+     * @key - Feature Key
+     * @type - Feature Type
+     * @description - Get a Int64 | Float | String | Binary value
+     */
+    public getFeature(recordPos: number, key: string, type: FeatureType): string | number | Uint8Array {
+        // Guard.expression(recordPos, (num) => num >= 0 && num < this.imageMessages.length);
```
mydiemho: remove

src/providers/export/tensorFlowRecords/tensorFlowReader.ts
```diff
@@ -62,4 +63,28 @@ export class TFRecordsReader {
...
+        const feature = message.getContext().getFeatureMap().get(key);
+
+        switch (type) {
+            case FeatureType.String:
+                return textDecode(feature.getBytesList().array[0][0]);
+            case FeatureType.Binary:
+                return feature.getBytesList().array[0][0];
+            case FeatureType.Int64:
+                return feature.getInt64List().array[0][0];
+            case FeatureType.Float:
```
mydiemho: is this needed when it's going to flow through to default anyway?

src/react/components/common/assetPreview/tfrecordAsset.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+import MockFactory from "../../../../common/mockFactory";
+import { TFRecordsBuilder, FeatureType } from "../../../../providers/export/tensorFlowRecords/tensorFlowBuilder";
+import HtmlFileReader from "../../../../common/htmlFileReader";
+
+describe("Image Asset Component", () => {
+    // tslint:disable-next-line:max-line-length
+    const dataImage64 = "R0lGODlhEAAQAMQAAORHHOVSKudfOulrSOp3WOyDZu6QdvCchPGolfO0o/XBs/fNwfjZ0frl3/zy7////wAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACH5BAkAABAALAAAAAAQABAAAAVVICSOZGlCQAosJ6mu7fiyZeKqNKToQGDsM8hBADgUXoGAiqhSvp5QAnQKGIgUhwFUYLCVDFCrKUE1lBavAViFIDlTImbKC5Gm2hB0SlBCBMQiB0UjIQA7";
+    const dataImage = new Uint8Array(Buffer.from(dataImage64, "base64"));
+
+    // tslint:disable-next-line:max-line-length
```
mydiemho: don't think this is needed anymore

src/common/htmlFileReader.test.ts
JacopoMangiavacchi: I think in general it's a good practice to always group tests around a similar scenario even when we have a single test in that category.  It's a good practice as you may want to add more tests on that category/describe later but in this specific case I don't see other possible tests there so I removed the describe call.

src/providers/export/tensorFlowRecords/tensorFlowHelpers.test.ts
JacopoMangiavacchi: I don't see the point here.  This string/array is only used once in this call.  Specially in test I see this more synthetic.

src/providers/export/tensorFlowRecords/tensorFlowHelpers.ts
JacopoMangiavacchi: Actually this is not an issue as in any language map on an empty list just do not start the enumeration.

src/providers/export/tensorFlowRecords/tensorFlowReader.ts
JacopoMangiavacchi: Good catch, the default it's not needed tks

src/react/components/common/assetPreview/tfrecordAsset.test.tsx
JacopoMangiavacchi: Unfortunately I still see lint error without it.

JacopoMangiavacchi: good catch tks

src/react/components/common/assetPreview/tfrecordAsset.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { IAssetProps } from "./assetPreview";
+import { IAsset, AssetType } from "../../../../models/applicationState";
+import HtmlFileReader from "../../../../common/htmlFileReader";
+import { TFRecordsReader } from "../../../../providers/export/tensorFlowRecords/tensorFlowReader";
+import { FeatureType } from "../../../../providers/export/tensorFlowRecords/tensorFlowBuilder";
+
+/**
+ * Properties for TFRecord Asset Image component
+ */
+export interface ITFRecordProps extends IAssetProps, React.ClassAttributes<TFRecordAsset> {
```
tbarlow12: Is there a reason to use this empty props as opposed to `IAssetProps`?

JacopoMangiavacchi: Not really, you're right it's empty interface.  I replicated same behavior as in imageAsset where it's there too even if empty but you're right.  Let me just remove this on both on the other PR.  Thanks

PR #556, opened 2019-02-11
Adding visited/tagged marks on video timeline AB #16772

tmiller-msft:
- Adding marks (yellow visited/green tagged) on video timeline
- Adding tests
- Cleaning up seeking of video to all go through same method, and adding new helpers
- Next/Previous "frame"
- Tests/strings

### DIFF
--- a/src/react/components/common/assetPreview/assetPreview.tsx
+++ b/src/react/components/common/assetPreview/assetPreview.tsx
@@ -47,8 +47,8 @@ export interface IAssetPreviewState {
  * @description - Small preview of assets for selection in editor page
  */
 export class AssetPreview extends React.Component<IAssetPreviewProps, IAssetPreviewState> {
-     /** Default properties for component if not defined */
-     public static defaultProps: IAssetPreviewProps = {
+    /** Default properties for component if not defined */
+    public static defaultProps: IAssetPreviewProps = {
         project: null,
         asset: null,
         childAssets: [],
--- a/src/react/components/common/assetPreview/assetPreview.tsx
+++ b/src/react/components/common/assetPreview/assetPreview.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { IAsset, AssetType } from "../../../../models/applicationState";
+import { IAsset, AssetType, IProject } from "../../../../models/applicationState";
 import { strings } from "../../../../common/strings";
 import { ImageAsset } from "./imageAsset";
 import { VideoAsset } from "./videoAsset";
@@ -10,6 +10,8 @@ export type ContentSource = HTMLImageElement | HTMLVideoElement;
  * AssetPreview component properties
  */
 export interface IAssetProps {
+    /** The current project */
+    project: IProject;
     /** The Asset to preview */
     asset: IAsset;
     /** The child assets (ex. video frames) of the parent asset */
@@ -45,8 +47,8 @@ export interface IAssetPreviewState {
  * @description - Small preview of assets for selection in editor page
  */
 export class AssetPreview extends React.Component<IAssetPreviewProps, IAssetPreviewState> {
-    /** Default properties for component if not defined */
     public static defaultProps: IAssetPreviewProps = {
+        project: null,
         asset: null,
         childAssets: [],
         autoPlay: false,
@@ -70,13 +72,15 @@ export class AssetPreview extends React.Component<IAssetPreviewProps, IAssetPrev
                     </div>
                 }
                 {asset.type === AssetType.Image &&
-                    <ImageAsset asset={parentAsset}
+                    <ImageAsset project={this.props.project}
+                        asset={parentAsset}
                         onLoaded={this.onAssetLoad}
                         onActivated={this.props.onActivated}
                         onDeactivated={this.props.onDeactivated} />
                 }
                 {(asset.type === AssetType.Video || asset.type === AssetType.VideoFrame) &&
-                    <VideoAsset asset={parentAsset}
+                    <VideoAsset project={this.props.project}
+                        asset={parentAsset}
                         childAssets={childAssets}
                         timestamp={asset.timestamp}
                         autoPlay={autoPlay}
--- a/src/react/components/common/assetPreview/assetPreview.tsx
+++ b/src/react/components/common/assetPreview/assetPreview.tsx
@@ -47,7 +47,8 @@ export interface IAssetPreviewState {
  * @description - Small preview of assets for selection in editor page
  */
 export class AssetPreview extends React.Component<IAssetPreviewProps, IAssetPreviewState> {
-    public static defaultProps: IAssetPreviewProps = {
+     /** Default properties for component if not defined */
+     public static defaultProps: IAssetPreviewProps = {
         project: null,
         asset: null,
         childAssets: [],
--- a/src/react/components/common/assetPreview/assetPreview.tsx
+++ b/src/react/components/common/assetPreview/assetPreview.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { IAsset, AssetType, IProject } from "../../../../models/applicationState";
+import { IAsset, AssetType, IProjectVideoSettings } from "../../../../models/applicationState";
 import { strings } from "../../../../common/strings";
 import { ImageAsset } from "./imageAsset";
 import { VideoAsset } from "./videoAsset";
@@ -10,12 +10,12 @@ export type ContentSource = HTMLImageElement | HTMLVideoElement;
  * AssetPreview component properties
  */
 export interface IAssetProps {
-    /** The current project */
-    project: IProject;
     /** The Asset to preview */
     asset: IAsset;
     /** The child assets (ex. video frames) of the parent asset */
     childAssets?: IAsset[];
+    /** Additional settings for this asset */
+    additionalSettings?: IAssetPreviewSettings;
     /** Event handler that fires when the asset has been loaded */
     onLoaded?: (ContentSource: ContentSource) => void;
     /** Event handler that fires when the asset has been activated (ex. Video resumes playing) */
@@ -43,13 +43,20 @@ export interface IAssetPreviewState {
 }
 
 /**
+ * Settings used by the various asset previews
+ * @member videoSettings - Video settings for this asset
+ */
+export interface IAssetPreviewSettings {
+    videoSettings: IProjectVideoSettings;
+}
+
+/**
  * @name - Asset Preview
  * @description - Small preview of assets for selection in editor page
  */
 export class AssetPreview extends React.Component<IAssetPreviewProps, IAssetPreviewState> {
     /** Default properties for component if not defined */
     public static defaultProps: IAssetPreviewProps = {
-        project: null,
         asset: null,
         childAssets: [],
         autoPlay: false,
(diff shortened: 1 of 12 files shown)


src/react/components/common/assetPreview/assetPreview.tsx
```diff
@@ -10,6 +10,8 @@ export type ContentSource = HTMLImageElement | HTMLVideoElement;
  * AssetPreview component properties
  */
 export interface IAssetProps {
+    /** The current project */
+    project: IProject;
```
wbreza: Instead of expected the whole object here I would like to scope this down a bit so only the asset level properties relevant for the asset preview.  @PIC123 has something in his current PR which should map to the properties that you need.

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -102,15 +115,17 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
     public componentDidUpdate(prevProps: Readonly<IVideoAssetProps>) {
         if (this.props.asset !== prevProps.asset) {
             this.setState({ loaded: false });
+        } else {
+            this.addAssetTimelineTags(this.props.childAssets, this.videoPlayer.current.getState().player.duration);
```
wbreza: Is this the right condition?  It feels like this should only run when `this.props.childAssets` changes?

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -102,15 +115,17 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
         }
         if (this.props.timestamp !== prevProps.timestamp) {
             this.seekToTimestamp();
         }
     }
 
     /**
-     * Bound to the "Next Tagged Frame" button
-     * Seeks the user to the next tagged video frame
+     * Bound to the "Previous Tagged Frame" button
```
wbreza: Thanks for fixing comments.

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -138,38 +153,97 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
-     * Seeks the video to the timestamp specified in the component properties
+     * @name - Move to the next expected frame
+     * @description - Moves the videos current position forward one frame based on the current
+     * project settings for frame rate extraction
      */
-    private seekToTimestamp = () => {
-        if (this.props.timestamp > 0) {
-            this.videoPlayer.current.pause();
-            this.videoPlayer.current.seek(this.props.timestamp);
+    private moveNextExpectedFrame = () => {
```
wbreza: I mentioned a similar comments in @PIC123's PR.  Should the sudo key frames be calculated from the absolute beginning of the video vs the current player time?  I don't know what the answer is but i know the frames would start getting offset if you manually skipped around in the video and then also utilized this skip to previous/next frame feature.

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -138,38 +153,97 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
+
+    /**
+     * @name - Move to nearest child frame
+     * @description - Move to the nearest tagged or visited frame from where the video's current
+     * position is, within a certain threshold
+     * @returns true if it moved position; false otherwise
+     */
+    private moveToNearestChildFrame(): boolean {
+        const maximumThreshold = 1; // one second threhsold
+        const timestamp = this.videoPlayer.current.getState().player.currentTime;
```
wbreza: Looks like we have quite a few places where we are calling `this.videoPlayer.current.getState()` now.  May we should make a tiny helper method that returns the current state mapped to our `IVideoPlayerState` interface.

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -225,4 +302,66 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
+     */
+    private addAssetTimelineTags(childAssets: any[], videoDuration: number) {
+        const innerHtml: string = this.getRenderedAssetTagLinesText(childAssets, videoDuration);
+        if (!this.timelineElement) {
+            // Find the progress holder so we can add the markers to that
+            const progressControls = document.getElementsByClassName("video-react-progress-control");
+            let progressHolderElement: Element = null;
+
+            // Find any that have an editor page content body type, this will be the one we use
+            Array.from(progressControls).forEach( (element) => {
```
wbreza: Any reason we can't do a standard DOM search?  ex) `document.querySelector('.editor-page-content-body')` or `myElement.querySelector('.editor-page-content-body')` if we already have a scoped element?

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -225,4 +302,66 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
+            const childPosition: number = (childAsset.timestamp / videoDuration);
+            tagTimeLines.push(<div key={childAsset.timestamp}
+                className = {childAsset.state === AssetState.Tagged ?
+                    "video-timeline-tagged" : "video-timeline-untagged"}
+                style={{
+                    left: (childPosition * 100) + "%",
+                 }} />);
+        }
+        const taggedAssetDiv = <div>{tagTimeLines}</div>;
+        return ReactDOMServer.renderToStaticMarkup(taggedAssetDiv);
```
wbreza: I've never used this `ReactDOMServer` before... how is it performing?  I'm curious what is the different between this and `ReacDOM.render(...)`.  If you take a look @ `/src/index.tsx` the root of the app is found and rendered with the later approach.  I'm thinking this may be closer to the recommended way of doing what you are attempting here - Creating your root element, then rendering in a bunch of components inside of it?

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -188,6 +262,9 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
                 this.props.onLoaded(this.videoPlayer.current.video.video);
             }
         });
+
+        // Once the video is loaded, add any asset timeline tags
```
wbreza: `onVideoStateChange` fires a lot... why do we need to keep re-rendering the tick marks?  Doesn't it only change if the child assets change?

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -138,38 +153,97 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
         if (!this.state.loaded && state.readyState === 4 && state.readyState !== prev.readyState) {
             // Video initial load complete
             this.raiseLoaded();
             this.raiseActivated();
             this.seekToTimestamp();
         } else if (state.paused && (state.currentTime !== prev.currentTime || state.seeking !== prev.seeking)) {
-            // Video is paused
-            this.raiseChildAssetSelected(state);
+            // Video is paused, first try to move to a nearby frame if one exists
+            if (!this.moveToNearestChildFrame()) {
```
wbreza: I'm curious to the end user experience of this change... If i'm attempting to manually scrub in a video to find the right frame to tag and I'm close to a previous tagged/visited frame it would automatically skip me back there?

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -225,4 +302,66 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
+     * @member videoDuration - Length (in seconds) of the video
+     */
+    private getRenderedAssetTagLinesText(childAssets: any[], videoDuration: number): string {
+        const tagTimeLines: any = [];
+
+        // Add some markers for frames that have been visited with yellow and tagged with green
+        for (const childAsset of childAssets) {
+            // Calcualte the left position
+            const childPosition: number = (childAsset.timestamp / videoDuration);
+            tagTimeLines.push(<div key={childAsset.timestamp}
```
wbreza: We should add click handlers on the lines so we don't need to hook into the seek behavior.

tmiller-msft: I was worried about having so many click handlers, but I can try this.

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: I argued with myself a lot about this. Without it, it's near impossible to sync to the "right" spot, but it also removes the ability for super fine grained scrubbing too. Maybe if the click handlers work that solves the problem.

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: I originally used ReactDOM.Render, but that seemed to simply replace the children rather than append them, but it's possible I was misunderstanding what "container" meant in that context, I will look into it.

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: I can see the argument both ways. I don't feel super strongly about either honestly, we should just make a call and do that.

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: Updated

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: Yeah, this should only update when childAssets changes, when I was doing my testing that was the only thing I was changing, so naturally it was the only thing I saw changing. :)

src/react/components/common/assetPreview/assetPreview.tsx
tmiller-msft: I'm fine with whatever limited scope is best. @PIC123 and my PR have some duplicates we should probably reconcile before the final merge

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: tbh, I thought they got messed up during my merge, so it was halfway on accident. :)

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: I'm not sure I understand this question? We only add the timeline marks on load and when the child assets change, it's never called in onVideoStateChange?

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: Updated

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: Updated, I had to switch no matter what, because the static markup wouldn't work with onClick events.

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: Removed since the div clicks seem to solve the problem this was intending, although this still can be a bit weird if you stop very near an already existing child asset.

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: Added

src/react/components/common/assetPreview/videoAsset.tsx
tmiller-msft: Updated to lock to video key frame rate per discussion

src/react/components/common/assetPreview/assetPreview.tsx
tmiller-msft: Updated to use additional settings

src/react/components/common/assetPreview/videoAsset.tsx
wbreza: Sorry, the github UI confused me 👍

src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -225,4 +288,89 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
+     */
+    private getRenderedAssetTagLinesElements(childAssets: any[], videoDuration: number) {
+        const tagTimeLines: any = [];
+
+        // Add some markers for frames that have been visited with yellow and tagged with green
+        for (const childAsset of childAssets) {
+            // Calcualte the left position
+            const childPosition: number = (childAsset.timestamp / videoDuration);
+            tagTimeLines.push(<div key={childAsset.timestamp}
+                onClick = {this.handleAssetTimelineClick}
```
wbreza: FYI, you can also specify an inline func to pass params ex. `onClick={() => this.handleAssetTimelineClick(childAsset.timestamp)}` which will retain the correct timestamp in the closure.

PR #557, opened 2019-02-11
WIP: Elhalper/import v1 project

elizabethhalper:
Adds the ability to import a V1 project to V2


src/common/htmlFileReader.test.ts
```diff
@@ -9,7 +9,7 @@ describe("Html File Reader", () => {
         const file = new File([blob], "test.txt");
 
         const actualText = await HtmlFileReader.readAsText(file);
-        expect(actualText).toEqual(expectedText);
+        expect(actualText.content).toEqual(expectedText);
```
wbreza: What causes this test to require this change?

src/common/htmlFileReader.ts
```diff
@@ -11,15 +11,20 @@ export default class HtmlFileReader {
      * Reads the file and returns the string value contained
      * @param file HTML file to read
      */
-    public static readAsText(file: File): Promise<string | ArrayBuffer> {
+    public static readAsText(file: File): Promise<any | ArrayBuffer | File> {
```
wbreza: Doesn't look like the return type matches up any more after this change.  I'd prefer to see an interface mapped to your new `fileInfo` object.

src/models/applicationState.ts
```diff
@@ -108,6 +110,62 @@ export interface IProject {
...
+    width: number;
+    height: number;
+    box: { [point: string]: number };
+    points: IPoint[];
+    UID: string;
+    id: number;
+    type: string;
+    tags: string[];
+    name: number;
+}
```
wbreza: Let's split out `v1` models into its own TS file

src/react/components/common/filePicker/filePicker.tsx
```diff
@@ -8,7 +8,7 @@ import HtmlFileReader from "../../../../common/htmlFileReader";
  * @member onError - Function to call on file picking error
  */
 interface IFilePickerProps {
-    onChange: (sender: SyntheticEvent, fileText: string | ArrayBuffer) => void;
+    onChange: (sender: SyntheticEvent, fileText: string | ArrayBuffer | File) => void;
```
wbreza: Similar to above - it looks like `fileText` here will always be your new `fileInfo` vs every returning only the string/arraybuffer.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -107,9 +119,19 @@ export default class HomePage extends React.Component<IHomepageProps> {
...
+                <Confirm title="Import Project"
+                    ref={this.importConfirm}
+                    message={(project: any) => `${strings.homePage.importProject.confirmation} ${project.file.name}
+                        ${strings.homePage.importProject.recommendation}`}
+                    confirmButtonColor="danger"
+                    onConfirm={this.convertProject} />
+                <Confirm title="Confirm Settings"
+                    ref={this.settingsConfirm}
+                    message={"Please confirm your new v2 project settings."}
+                    onConfirm={this.deleteProject} />
```
wbreza: Delete project on confirming setting?  By design?

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -122,12 +144,21 @@ export default class HomePage extends React.Component<IHomepageProps> {
...
+            projectJson = JSON.parse(project.content);
         } catch (error) {
             throw new AppError(ErrorCode.ProjectInvalidJson, "Error parsing JSON");
         }
 
-        await this.loadSelectedProject(projectJson);
+        // need a better check to tell if its v1
+        if (projectJson.name == null) {
+            try {
+                await this.importConfirm.current.open(project);
```
wbreza: I have an idea to make this easier... I can return a promise from the `.open(...)` call that returns a value indicating whether or not the user confirmed the dialog.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -139,11 +170,31 @@ export default class HomePage extends React.Component<IHomepageProps> {
     }
 
     private loadSelectedProject = async (project: IProject) => {
-        await this.props.actions.loadProject(project);
-        this.props.history.push(`/projects/${project.id}/edit`);
+        console.log(project);
```
wbreza: At this point the incoming project has already been converted and the confirmation should have already happened. Also - any need to bring the user to the project settings page vs. the edit page in this case?  The project has already been saved and they can easily go back to the project settings to make any additional changes.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -139,11 +170,31 @@ export default class HomePage extends React.Component<IHomepageProps> {
...
         await this.props.actions.deleteProject(project);
     }
+
+    private convertProject = async (project: any) => {
+        const importService = new ImportService();
+        let projectJson;
+        try {
+            projectJson = await importService.convertV1(project);
+        } catch (e) {
+            throw new AppError(ErrorCode.ProjectUploadError, "Error uploading v1 project file");
```
wbreza: Please create a new `ErrorCode` for this condition since it's very different from an project upload error.

src/react/components/pages/projectSettings/projectSettingsPage.tsx
```diff
@@ -54,6 +53,8 @@ export default class ProjectSettingsPage extends React.Component<IProjectSetting
         const projectId = this.props.match.params["projectId"];
         if (!this.props.project && projectId) {
             const project = this.props.recentProjects.find((project) => project.id === projectId);
+            console.log(project);
+            this.props.applicationActions.ensureSecurityToken(this.props.appSettings, project);
```
wbreza: The project should already be saved and have a security token at this point.

src/react/components/pages/projectSettings/projectSettingsPage.tsx
```diff
@@ -85,7 +86,7 @@ export default class ProjectSettingsPage extends React.Component<IProjectSetting
     private onFormSubmit = async (project: IProject) => {
         const isNew = !(!!project.id);
 
-        await this.ensureSecurityToken(project);
+        await this.props.applicationActions.ensureSecurityToken(this.props.appSettings, project);
```
wbreza: Instead of loading in the `appSettings` into this component we can have the redux action lookup that appSettings state during this action.

src/redux/actions/applicationActions.ts
```diff
@@ -50,6 +53,38 @@ export function saveAppSettings(appSettings: IAppSettings): (dispath: Dispatch)
     };
 }
 
+/**
+ * Ensures that a valid security token is associated with the project, otherwise creates one
+ * @param project The project to validate
+ */
+export function ensureSecurityToken(appSettings: IAppSettings, project: IProject):
```
wbreza: Mentioned above, but let's have this action lookup the appSettings so we don't need to pass it in here.

src/redux/actions/projectActions.ts
```diff
@@ -59,7 +61,7 @@ export function saveProject(project: IProject):
         const projectService = new ProjectService();
 
         if (projectService.isDuplicate(project, appState.recentProjects)) {
-            throw new Error(`Project with name '${project.name}
+            throw new AppError(ErrorCode.ProjectDuplicateName, `Project with name '${project.name}
```
wbreza: 👍

src/services/importService.ts
```diff
@@ -0,0 +1,186 @@
...
+// const TagColors = require("../react/components/common/tagsInput/tagColors.json");
+
+/**
+ * Functions required for an import service
+ * @member convertV1 - Converts a v1 project to v2 project
+ * @member generateAssets - Generates v2 assets based on v1 project file
+ * @member generateConnections - Generates v2 connections based on location of v1 project file
+ */
+export interface IImportService {
+    convertV1(project: any): Promise<IProject>;
```
wbreza: Since you have your `v1` models you should set the types here vs using `any`

src/services/importService.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/**
+ * @name - Import Service
+ * @description - Functions for importing v1 projects to v2 application
+ */
+// change any to { content, file } object eventually
+export default class ImportService implements IImportService {
+    public convertV1(project: any): Promise<IProject> {
+        return new Promise<IProject>((resolve, reject) => {
```
wbreza: You should be able to use standard `async/await` here vs having to wrap this in a promise.

src/services/importService.ts
```diff
@@ -0,0 +1,186 @@
...
+            let originalProject: IV1Project;
+            let convertedProject: IProject;
+            let connections: IConnection[];
+            let parsedTags: ITag[];
+            let generatedAssetMetadata: IAssetMetadata[];
+
+            originalProject = JSON.parse(project.content);
```
wbreza: Are we confident the JSON is valid at this point?  If not we should try/catch this.

src/services/importService.ts
```diff
@@ -0,0 +1,186 @@
...
+            resolve(convertedProject);
+        });
+    }
+
+    /**
+     * Generate connections from v1 project file location
+     * @param project - V1 Project Content and File Information
+     */
+    private generateConnections(project: any): IConnection[] {
+        const sourceConnection: IConnection = {
```
wbreza: Let's rename the connection to something memorable based on the project file name.

src/services/importService.ts
```diff
@@ -0,0 +1,186 @@
...
+
+        for (const frameName in originalProject.frames) {
+            if (originalProject.frames.hasOwnProperty(frameName)) {
+                const frameRegions = originalProject.frames[frameName];
+                assetState = originalProject.visitedFrames.indexOf(frameName) > -1 && frameRegions.length > 0
+                             ? AssetState.Tagged : (originalProject.visitedFrames.indexOf(frameName) > -1
+                             ? AssetState.Visited : AssetState.NotVisited);
+
+                assetMetadata = {
+                    asset: {
```
wbreza: We have a method on the `AssetService` to create an asset from a file path.  We should re-use this as much as possible to make sure everything is inline.  Then, once you receive the asset you can then override any specific properties as needed based on the current state of the asset from the v1 project.

src/services/importService.ts
```diff
@@ -0,0 +1,186 @@
...
+                        size: {
+                            width: frameRegions.length > 0 ? frameRegions[0].width : null,
+                            height: frameRegions.length > 0 ? frameRegions[0].height : null,
+                        },
+                        format: frameName.split(".").pop(),
+                    },
+                    regions: [],
+                };
+
+                for (const region of frameRegions) {
```
wbreza: Also - sync with @tbarlow12  - he may have some helper methods that help convert from v1 regions to v2 regions... since I believe this may be very similar if not the same as the region data shares in canvas tools.

src/services/projectService.ts
```diff
@@ -53,6 +53,10 @@ export default class ProjectService implements IProjectService {
                     project.id = shortid.generate();
                 }
 
+                if (project.version !== "v2") {
```
wbreza: We should pull the exact version info from the executing app.

wbreza: We should also add version to the `IProject` interface as required.

src/common/htmlFileReader.test.ts
elizabethhalper: I am now having the reader return both the contents of the file and the file information. This is because the V1 project format doesn't include any information about the file itself. I'm using that info to get the name and the path, which is necessary for creating a V2 project as well as saving the assets in the proper place

src/common/htmlFileReader.ts
elizabethhalper: So you want me to create an IFileInfo interface in applicationState for this return type? That may also be useful later on when I am loading the project information to my service. I have "any" right now, but wanted to change it to something more specific.

src/react/components/pages/homepage/homePage.tsx
elizabethhalper: No, this was a mistake. I've changed this to a message box with no onConfirm method

src/react/components/pages/homepage/homePage.tsx
elizabethhalper: We had discussed earlier bringing them straight to the settings page so they can verify the info, given I've generated the titles of the project and its connections. But if we want to go straight to the edit page, I'd be fine with that and can get rid of the if clause

src/services/importService.ts
elizabethhalper: Already did :) I'll push that up now

src/services/projectService.ts
elizabethhalper: What do you mean by your first comment? And if it is required, where would we set the version when the project is created?

src/react/components/pages/homepage/homePage.tsx
elizabethhalper: Do you mean in order to check if it is a v1 file? If this pops up every time the user tries to open a file, wouldn't that be bad for the user, especially if they have mostly v2 projects?

src/services/importService.ts
elizabethhalper: Doesn't any async method have to return a promise?

src/services/importService.ts
elizabethhalper: When I leverage the asset service, it seems to be creating an asset based on the project json file, not the jpg's in that directory, and thus creating only one asset file. Would there be a reason for this?

src/react/components/pages/homepage/homePage.tsx
wbreza: This was just a note for myself to improve the experience for opening a modal that could return a promise vs. today it just returns void.

src/react/components/pages/homepage/homePage.tsx
wbreza: We're creating new files anyway - so if there was an issue we already saved the file.  They can always re-open the v1 file vs. the newly generated v2 file.  I would keep the experience the same as v2.   If we encounter any errors/warnings during the conversion we should till display a modal with this failures.

src/services/importService.ts
wbreza: Yes, async methods return a promise... but if your method is making other async calls using await then your return value will already be wrapped in a promise.

src/services/projectService.ts
wbreza: For example, if we release a new version `2.1.5` as defined withing the `package.json` then our app should use this version to auto-version the project file anytime it is saved.

PR #558, opened 2019-02-11
Cut, Copy, Paste and Clear

tbarlow12:
- Introduced `CanvasHelpers` to provide calculations for the Canvas component
- Added button stubs, but no functionality yet (coming soon in future PR)
- Hot keys:
   - `Ctrl + C` -> Copy
   - `Ctrl + X` -> Cut
   - `Ctrl + V` -> Paste
   - `Ctrl + A` -> Select all
   - `Ctrl + D` -> Clear


src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -40,8 +43,12 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
         contentSource: null,
         selectedRegions: [],
         canvasEnabled: true,
+        multiSelect: false,
     };
 
+    private clipBoard: ClipBoard<IRegion[]> = new ClipBoard<IRegion[]>();
+
+    private videoPlayer: React.RefObject<Player> = React.createRef<Player>();
```
wbreza: The video player does not exist directly in the canvas anymore and we shouldn't need this.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -96,10 +141,41 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
      */
     public onTagClicked = (tag: ITag) => {
         for (const region of this.state.selectedRegions) {
-            this.toggleTagOnRegion(region, tag);
+            CanvasHelpers.toggleTag(region.tags, tag);
+            this.editor.RM.updateTagsById(region.id, CanvasHelpers.getTagsDescriptor(region));
         }
     }
 
+    public onTagShiftClicked = (tag: ITag) => {
```
wbreza: Are you not using these?  If so, please remove.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -96,10 +141,41 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
+        console.log(`Shift clicked "${tag.name}" tag`);
+    }
+
+    public onTagCtrlShiftClicked = (tag: ITag) => {
+        console.log(`Ctrl shift clicked "${tag.name}" tag`);
+    }
+
+    /**
```
wbreza: js docs don't match method signature

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -96,10 +141,41 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
+     * Method called when deleting a region from the editor
+     * @param {string} id the id of the deleted region
+     * @param {boolean} multiselection boolean whether multiselect is active
+     * @returns {void}
+     */
+    public onRegionSelected = (id: string) => {
```
wbreza: Can this be private?

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -166,32 +169,23 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
-                            this.props.project.tags.find((t) => t.name === tempTag.name).color))));
-                } else {
-                    this.canvas.current.editor.RM.updateTagsById(selectedRegion.id, null);
-                }
-
-                return region;
-            });
-        }
-        this.onAssetMetadataChanged(selectedAsset);
+    public onTagClicked = (tag: ITag) => {
```
wbreza: Instead of having the `editor page` call into the `canvas` component to perform tag related operations I would like to see this model inverted where the `editor page` only sets up some internal state like `selectedTags` or similar which is then passed through props into the `canvas` where the `canvas` can then do the right thing based on its state.

wbreza: Let's discuss.

src/registerToolbar.ts
```diff
@@ -56,6 +51,38 @@ export default function registerToolbar() {
         type: ToolbarItemType.State,
     });
 
+    ToolbarItemFactory.register(CopyRegions, {
```
wbreza: Right now there are basically 2 categories of toolbar buttons... 1 where they have code that perform something onclick, and the other where their type is inspected in the editor page to dispatch operations.  For these they appear to do the later, but I think there is some code bloat going on where we can simplify this by creating a new `GenericToolbarButton` that doesn't really do anything and avoids us having to create these stub button components.

src/common/clipboard.ts
```diff
@@ -0,0 +1,20 @@
+export class ClipBoard<T> {
```
wbreza: We should look to standardize the clipboard usage based on the working draft of the W3C clipboard API @ https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

I would suggest we use the standard API but fall back to a polyfill if not supported by the current browser.

src/registerToolbar.ts
```diff
@@ -56,6 +51,38 @@ export default function registerToolbar() {
...
+        name: "copyRegions",
```
wbreza: Shouldn't the accelerators be hooked up here?

src/react/components/common/tagsInput/tagsInput.tsx
```diff
@@ -159,7 +160,10 @@ export default class TagsInput<T extends ITagsInputProps> extends React.Componen
         if (event.target.lastChild) {
             return event.target.lastChild.data;
         }
-        return (event.target.innerText || event.currentTarget.innerText).trim();
+        return (event.target.innerText || event.currentTarget.innerText)
```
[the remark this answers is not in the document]
tbarlow12: So store something like `[9] Tag9` in the `ITag` object?

src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -13,24 +13,32 @@ import { Editor } from "vott-ct/lib/js/CanvasTools/CanvasTools.Editor";
...
+                             includeKeyboardManager = false)
         : ReactWrapper<ICanvasProps, ICanvasState, Canvas> {
-        return mount(
-            <Canvas {...canvasProps}>
-                <AssetPreview {...assetPreviewProps} />
-            </Canvas>,
-        );
+            if (includeKeyboardManager) {
+                return mount(
+                    <KeyboardManager>
```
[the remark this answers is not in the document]
tbarlow12: Maybe I misunderstood what we discussed. I thought I needed to wrap the component in KeyboardManager similar to [how it is done in the editor page](https://github.com/Microsoft/VoTT/blob/v2/src/react/components/pages/editorPage/editorPage.test.tsx#L39) in order to test the key commands

src/react/components/pages/editorPage/canvas.tsx
tbarlow12: This was previously public, and I think it's to accommodate for [testing ](https://github.com/Microsoft/VoTT/blob/v2/src/react/components/pages/editorPage/canvas.test.tsx#L175)

src/react/components/pages/editorPage/editorPage.tsx
tbarlow12: Agreed. I'll re-work this and get back to you

src/react/components/pages/editorPage/canvas.tsx
JacopoMangiavacchi: In mocha you can inherit your test class to extend the class you want to test so you have access to private/protected method.  Is there anything similar in jest ?

Also Docs mentioned method to "delete" vs "select"

src/react/components/pages/editorPage/editorPage.tsx
tbarlow12: So the only thing in my mind that's keeping some of this in the canvas is the `selectedTags` issue. The `onTagClicked` needs to update all selected tags

tbarlow12: Oh wait, I think I know what to do here. Could help with the tag locking as well. I'll get back to you

src/registerToolbar.ts
tbarlow12: Would you have accelerators on both the buttons and within the `Canvas` component?

tbarlow12:
Closing in favor of smaller PRs

PR #564, opened 2019-02-12
Update project when in-use connection changes AB#16921

wbreza:
- When you edit a connection settings while having a project loaded it will update the project to use the updated connection settings
- Optimized data stored in project / asset files

### DIFF
--- a/src/common/mockFactory.ts
+++ b/src/common/mockFactory.ts
@@ -2,7 +2,7 @@ import shortid from "shortid";
 import {
     AssetState, AssetType, IApplicationState, IAppSettings, IAsset, IAssetMetadata,
     IConnection, IExportFormat, IProject, ITag, StorageType, ISecurityToken,
-    EditorMode, IAppError, IProjectVideoSettings, AppError, ErrorCode,
+    EditorMode, IAppError, IProjectVideoSettings, ErrorCode,
     IPoint, IRegion, RegionType, IBoundingBox,
 } from "../models/applicationState";
 import { ExportAssetState } from "../providers/export/exportProvider";
@@ -163,8 +163,8 @@ export default class MockFactory {
 
     /**
      * Creates array of fake IAsset
-     * @param count Number of assets to create
-     * @param videoFirst true if the first asset should be video; false otherwise
+     * @param count Number of assets to create (default: 10)
+     * @param startIndex The index that the assets should start at (default: 1)
      */
     public static createTestAssets(count: number = 10, startIndex: number = 1): IAsset[] {
         const assets: IAsset[] = [];
--- a/src/common/mockFactory.ts
+++ b/src/common/mockFactory.ts
@@ -166,14 +166,14 @@ export default class MockFactory {
      * @param count Number of assets to create
      * @param videoFirst true if the first asset should be video; false otherwise
      */
-    public static createTestAssets(count: number = 10, videoFirst: boolean = true): IAsset[] {
+    public static createTestAssets(count: number = 10, startIndex: number = 1): IAsset[] {
         const assets: IAsset[] = [];
-        for (let i = 1; i <= count; i++) {
-            assets.push((i % 2 === 1) ?
-                videoFirst ? MockFactory.createVideoTestAsset(i.toString()) :
-                    MockFactory.createTestAsset(i.toString()) :
-                !videoFirst ? MockFactory.createVideoTestAsset(i.toString()) :
-                    MockFactory.createTestAsset(i.toString()));
+        for (let i = startIndex; i < (count + startIndex); i++) {
+            const newAsset = (i % 2 === 1)
+                ? MockFactory.createVideoTestAsset(i.toString())
+                : MockFactory.createTestAsset(i.toString());
+
+            assets.push(newAsset);
         }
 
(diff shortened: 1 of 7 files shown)


src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -322,14 +321,18 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
 
         this.loadingProjectAssets = true;
 
-        await this.props.actions.loadAssets(this.props.project);
-        const assets = _.values(this.props.project.assets);
+        const projectAssets = _.values(this.props.project.assets);
```
wbreza: Project file has been updated to only store assets that have been visited/tagged.  This change will merge in project file assets with the assets from the asset/storage provider.

src/redux/reducers/currentProjectReducer.ts
```diff
@@ -21,34 +21,32 @@ export const reducer = (state: IProject = null, action: AnyAction): IProject =>
             return null;
         case ActionTypes.LOAD_PROJECT_SUCCESS:
             return { ...action.payload };
-        case ActionTypes.LOAD_PROJECT_ASSETS_SUCCESS:
```
wbreza: When loading assets from the asset provider it no longer saves them in the project assets by default.  This is a perf optimization in the event of a large number of assets to help reduce project load time.

src/redux/reducers/currentProjectReducer.ts
```diff
@@ -21,34 +21,32 @@ export const reducer = (state: IProject = null, action: AnyAction): IProject =>
...
-                };
-            } else {
+            const updatedAssets = { ...state.assets } || {};
+            updatedAssets[action.payload.asset.id] = { ...action.payload.asset };
+
+            return {
+                ...state,
+                assets: updatedAssets,
+            };
+        case ActionTypes.SAVE_CONNECTION_SUCCESS:
```
wbreza: If the connection being updated is in use by the current project then also update the current project source/target connection

src/services/assetService.ts
```diff
@@ -149,8 +149,12 @@ export class AssetService {
     public async save(metadata: IAssetMetadata): Promise<IAssetMetadata> {
         Guard.null(metadata);
 
-        const fileName = `${metadata.asset.id}${constants.assetMetadataFileExtension}`;
-        await this.storageProvider.writeText(fileName, JSON.stringify(metadata, null, 4));
+        // Only save asset metadata if asset is in a tagged state
```
wbreza: Perf improvement: No need to store an asset metadata JSON file if the asset has not been tagged - otherwise the core asset info is already going to be inside the project file.

src/common/mockFactory.ts
```diff
@@ -166,14 +166,14 @@ export default class MockFactory {
...
+    public static createTestAssets(count: number = 10, startIndex: number = 1): IAsset[] {
         const assets: IAsset[] = [];
-        for (let i = 1; i <= count; i++) {
-            assets.push((i % 2 === 1) ?
-                videoFirst ? MockFactory.createVideoTestAsset(i.toString()) :
-                    MockFactory.createTestAsset(i.toString()) :
-                !videoFirst ? MockFactory.createVideoTestAsset(i.toString()) :
-                    MockFactory.createTestAsset(i.toString()));
+        for (let i = startIndex; i < (count + startIndex); i++) {
+            const newAsset = (i % 2 === 1)
```
mydiemho: I support this easier-to-read coding style

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -138,12 +142,12 @@ describe("Editor Page Component", () => {
             name: testProject.name,
         };
 
-        const expectedAssetMetadtata: IAssetMetadata = getMockAssetMetadata(testAssets);
-
         await MockFactory.flushUi();
 
+        const expectedAssetMetadtata: IAssetMetadata = getMockAssetMetadata(projectAssets);
```
mydiemho: what prompted the order of operation change?

src/redux/reducers/currentProjectReducer.test.ts
```diff
@@ -39,14 +40,35 @@ describe("Current Project Reducer", () => {
...
+        const currentProject = MockFactory.createTestProject("1");
+        const state: IProject = currentProject;
+
+        const sourceConnection = {...currentProject.sourceConnection};
+        sourceConnection.description += "updated";
+
+        const action = saveConnectionAction(sourceConnection);
+        const result = reducer(state, action);
+
+        expect(result.sourceConnection).toEqual(action.payload);
```
mydiemho: should we also test that current project has change?

src/react/components/pages/editorPage/editorPage.test.tsx
wbreza: Everything was working fine, just my OCD since all the expectations happen after the UI has been updated.

src/redux/reducers/currentProjectReducer.test.ts
wbreza: Updated expectations.

PR #566, opened 2019-02-12
Import region metadata from TFRecords

JacopoMangiavacchi:


src/providers/export/tensorFlowRecords.ts
```diff
@@ -183,10 +183,12 @@ export class TFRecordsJsonExportProvider extends ExportProvider<ITFRecordsJsonEx
 
                                         imageInfo.text.push(tag.name);
                                         imageInfo.label.push(index);
-                                        imageInfo.xmin.push(region.boundingBox.left);
-                                        imageInfo.ymin.push(region.boundingBox.top);
-                                        imageInfo.xmax.push(region.boundingBox.left + region.boundingBox.width);
-                                        imageInfo.ymax.push(region.boundingBox.top + region.boundingBox.height);
+                                        imageInfo.xmin.push(region.boundingBox.left / imageInfo.width);
```
wbreza: Why dividing by width/height here?

src/react/components/common/assetPreview/tfrecordAsset.test.tsx
```diff
@@ -0,0 +1,74 @@
+import React from "react";
+import { IAssetProps } from "./assetPreview";
+import { ReactWrapper, mount } from "enzyme";
+import { TFRecordAsset } from "./tfrecordAsset";
+import MockFactory from "../../../../common/mockFactory";
+import { TFRecordsBuilder, FeatureType } from "../../../../providers/export/tensorFlowRecords/tensorFlowBuilder";
+import HtmlFileReader from "../../../../common/htmlFileReader";
+
+describe("Image Asset Component", () => {
```
wbreza: Nit: Update description for test block

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -308,6 +308,7 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
 
         this.setState({
             selectedAsset: assetMetadata,
+            assets: _.values(this.props.project.assets),
```
wbreza: Remove this line - its no longer needed.

src/services/assetService.ts
```diff
@@ -1,10 +1,27 @@
...
 import { AssetProviderFactory, IAssetProvider } from "../providers/storage/assetProviderFactory";
 import { StorageProviderFactory, IStorageProvider } from "../providers/storage/storageProviderFactory";
 import { constants } from "../common/constants";
+import HtmlFileReader from "../common/htmlFileReader";
+import { TFRecordsReader } from "../providers/export/tensorFlowRecords/tensorFlowReader";
+import { FeatureType } from "../providers/export/tensorFlowRecords/tensorFlowBuilder";
+// tslint:disable-next-line:no-var-requires
+const TagColors = require("../react/components/common/tagsInput/tagColors.json");
+
+interface ITFRecordObjectArray {
```
wbreza: It doesn't feel like this interface belongs in this generic `assetService` file

src/services/assetService.ts
```diff
@@ -171,10 +190,74 @@ export class AssetService {
...
+                id: shortid.generate(),
+                type: RegionType.Rectangle,
+                tags: [{
+                    name: objectArray.textArray[index],
+                    color: TagColors[tagPos],
+                }],
+                boundingBox: {
+                    left: objectArray.xminArray[index] * objectArray.width,
+                    top: objectArray.yminArray[index] * objectArray.height,
+                    width: (objectArray.xmaxArray[index] - objectArray.xminArray[index]) * objectArray.width,
```
wbreza: Why multiply? isn't width equal to `max - min` ?

src/providers/export/tensorFlowRecords.ts
JacopoMangiavacchi: I just figure it out TFRecords format is expecting relative points expressed in 0.0<=1.0 range respect to the image width and height

src/services/assetService.ts
JacopoMangiavacchi: I used tuple first but unfortunately TS/JS doesn't support named tuple properties and using the array syntax was really really bad to see in the code so I decided to use a JS object and as we're on TS I created this interface.

I agree it's not ideal to have this here but indeed this interface it's just needed for a private function in the assetService.

I can move the interface on a different file, as well as the getTFRecordObjectArrays() method that return this interface/object but I don't really want to put these in any TFRecords functions that I already have (like Builder, Reader, Helpers) as this code is specific to the way VoTT use TFRecords and not about how TFRecord internal implementation works.  This is specially true if we want to ship the generic TFRecords stuff on a separated npm package.

Do you think it's better to create this separated file for this interface and method and store it in the same services folder ?  I'm not 100% sure it's better.

src/services/assetService.ts
JacopoMangiavacchi: Same reason as above.  The way we (v1) store these points in TFRecords format is in terms of relative points expressed in 0.0<=1.0 range respect to the image width and height

src/providers/export/tensorFlowRecords.ts
wbreza: OK - makes sense then

src/services/assetService.ts
wbreza: OK - Let's move it into our core models in `applicationState`.  Let's also rename to make it more clear... `ITFRecordObjectArray` has array in the name but its not an array, but rather has members which some members are an array.  Maybe just `ITFRecord` ?

JacopoMangiavacchi: ITFRecordMetadata now, in applicationState

src/providers/export/tensorFlowRecords/tensorFlowReader.ts
```diff
@@ -85,4 +85,27 @@ export class TFRecordsReader {
...
     }
+
+    /**
+     * @recordPos - Record Position
+     * @key - Feature Key
+     * @type - Feature Type
+     * @description - Get an array of Int64 | Float | String | Binary value
+     */
+    public getArrayFeature(recordPos: number, key: string, type: FeatureType): string[] | number[] | Uint8Array[] {
+        // Guard.expression(recordPos, (num) => num >= 0 && num < this.imageMessages.length);
```
PIC123: Nit: Remove unused commented code

src/react/components/common/assetPreview/tfrecordAsset.test.tsx
```diff
@@ -6,10 +6,12 @@ import MockFactory from "../../../../common/mockFactory";
 import { TFRecordsBuilder, FeatureType } from "../../../../providers/export/tensorFlowRecords/tensorFlowBuilder";
 import HtmlFileReader from "../../../../common/htmlFileReader";
 
-describe("Image Asset Component", () => {
+describe("TFRecord Asset Component", () => {
     // tslint:disable-next-line:max-line-length
     const dataImage64 = "R0lGODlhEAAQAMQAAORHHOVSKudfOulrSOp3WOyDZu6QdvCchPGolfO0o/XBs/fNwfjZ0frl3/zy7////wAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACH5BAkAABAALAAAAAAQABAAAAVVICSOZGlCQAosJ6mu7fiyZeKqNKToQGDsM8hBADgUXoGAiqhSvp5QAnQKGIgUhwFUYLCVDFCrKUE1lBavAViFIDlTImbKC5Gm2hB0SlBCBMQiB0UjIQA7";
     const dataImage = new Uint8Array(Buffer.from(dataImage64, "base64"));
+
+    // tslint:disable-next-line:max-line-length
```
PIC123: Nit: I don't think you need this here, the next line is quite short

src/react/components/common/assetPreview/tfrecordAsset.test.tsx
```diff
@@ -42,15 +44,15 @@ describe("Image Asset Component", () => {
         return mount(<TFRecordAsset {...props} />);
     }
 
-    function sleep(ms) {
-        return new Promise((resolve) => setTimeout(resolve, ms));
+    function wait() {
+        return new Promise((resolve) => setImmediate(resolve));
```
PIC123: Does this just wait for all processes to finish? Is this not like just using setImmediate normally?

JacopoMangiavacchi: They're not really the same. There're two different queues for io and functions where the second is on top of the other.  setImmediate go directly on the io queue.  From my point of view I reaally don't see a big differences in the browser but wallace suggest me to use setimmediate

PR #567, opened 2019-02-12
Update canvastools and fix tests

PIC123:
Updates vott-ct library and updates tests

### DIFF
--- a/src/react/components/pages/editorPage/canvas.test.tsx
+++ b/src/react/components/pages/editorPage/canvas.test.tsx
@@ -12,7 +12,7 @@ import { Editor } from "vott-ct/lib/js/CanvasTools/CanvasTools.Editor";
 
 jest.mock("vott-ct/lib/js/CanvasTools/Region/RegionsManager");
 import { RegionsManager } from "vott-ct/lib/js/CanvasTools/Region/RegionsManager";
-import { SelectionMode } from "vott-ct/lib/js/CanvasTools/Selection/AreaSelector";
+import { SelectionMode, AreaSelector } from "vott-ct/lib/js/CanvasTools/Selection/AreaSelector";
 
 describe("Editor Canvas", () => {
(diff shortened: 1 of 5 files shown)


src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -58,7 +58,8 @@ describe("Editor Canvas", () => {
         const editorMock = Editor as any;
         editorMock.prototype.addContentSource = jest.fn(() => Promise.resolve());
         editorMock.prototype.scaleRegionToSourceSize = jest.fn((regionData: any) => regionData);
-        editorMock.prototype.RM = new RegionsManager(null, null, null);
+        editorMock.prototype.RM = new RegionsManager(null, null);
```
mydiemho: not sure if this from a library or our code, but using abbreviations like these, `RM` & `AS`, makes it really hard to read the code

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -141,7 +141,7 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
      * @param {RegionData} regionData the RegionData of moved region
      * @returns {void}
      */
-    private onRegionMove = (id: string, regionData: RegionData) => {
+    private onRegionMoveEnd = (id: string, regionData: RegionData) => {
```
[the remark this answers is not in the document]
PIC123: So the issue is that onREgionMove initially got triggered for every tiny of the region, not just when the user finished moving the region. The end here ties into the onREgionMoveEnd event from canvastools that is called once the movement is done.

src/react/components/pages/editorPage/canvas.test.tsx
PIC123: This is from the canvastools library that we are using

PR #568, opened 2019-02-12
Fixed delete project bug AB#17024

wbreza:
- Fixes issue deleting project

Issue: After security tokens were introduced the storage provider was not able to be instantiated because it was attempting to load the provider from the encrypted provider settings.

### DIFF
--- a/src/react/components/pages/homepage/homePage.tsx
+++ b/src/react/components/pages/homepage/homePage.tsx
@@ -144,6 +144,10 @@ export default class HomePage extends React.Component<IHomepageProps> {
     }
 
     private deleteProject = async (project: IProject) => {
-        await this.props.actions.deleteProject(project);
+        try {
+            await this.props.actions.deleteProject(project);
+        } catch (error) {
+            throw new AppError(ErrorCode.ProjectDeleteError, "Error deleting project file");
+        }
     }
 }
--- a/src/redux/actions/projectActions.ts
+++ b/src/redux/actions/projectActions.ts
@@ -52,8 +52,8 @@ export function loadProject(project: IProject):
  * Dispatches Save Project action and resolves with IProject
  * @param project - Project to save
  */
-export function saveProject(project: IProject):
-    (dispatch: Dispatch, getState: () => IApplicationState) => Promise<IProject> {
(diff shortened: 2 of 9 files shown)


src/redux/actions/projectActions.ts
```diff
@@ -84,11 +84,24 @@ export function saveProject(project: IProject):
...
+    : (dispatch: Dispatch, getState: () => IApplicationState) => Promise<void> {
+    return async (dispatch: Dispatch, getState: () => IApplicationState) => {
+        const appState = getState();
         const projectService = new ProjectService();
-        await projectService.delete(project);
-        dispatch(deleteProjectAction(project));
+
+        // Lookup security token used to decrypt project settings
+        const securityToken = appState.appSettings.securityTokens
+            .find((st) => st.name === project.securityToken);
```
mydiemho: I prefer `securityToken` to `st`

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -144,6 +144,10 @@ export default class HomePage extends React.Component<IHomepageProps> {
     }
 
     private deleteProject = async (project: IProject) => {
-        await this.props.actions.deleteProject(project);
+        try {
+            await this.props.actions.deleteProject(project);
+        } catch (error) {
+            throw new AppError(ErrorCode.ProjectDeleteError, "Error deleting project file");
```
mydiemho: what happens if project file is gone but it's still showing in `Recent Projects`?  Is there a force delete option?

wbreza: I was also thinking that as well.  I didn't want to complicate this story with too much extra but i will add an item on the backlog to address this.

src/redux/actions/projectActions.ts
wbreza: I agree, but didn't want to have 2 scoped variables with the same name since `securityToken` is defined above.  I will consider renaming both of them.

src/react/components/pages/homepage/homePage.tsx
wbreza: AB#17063

PR #569, opened 2019-02-12
Fix bug applying tags when clicking on color bubble AB#16923

wbreza:
- Added `data-tag-name` data attribute to set the original tag name/key
- Updated event handling to use `currentTarget` so that the element will always be the same as the element that has the `click` handler.


wbreza:
@luisamiranda - We should get this change migrated into the `vott-react` project as well.

luisamiranda:
@wbreza This file is deleted by the integration PR, so these changes should only be on vott-react, right?

wbreza:
@luisamiranda - Yes, one your PR goes it it will be deleted.  Until that happens I would like to get this bug fixed because it is a poor user experience.

PR #574, opened 2019-02-13
Added CI script for updating complexity reports, pushing to blob

pjlittle:
We're moving from GitHub pages to Azure Blob Storage + Static Website for hosting our complexity analysis reports. After going down a few dead ends, found a relatively simple way to integrate this into our CI progress, using Azure CLI. 

Blob Storage will always be source-of-truth. Tested this locally on OSX and Linux (Ubuntu). 

Relevant links:
V2: https://vottv2.z5.web.core.windows.net/
V1: https://vottv1.z5.web.core.windows.net/

- Added CI script to fetch, update and push complexity reports
- Updated docs

Once the Azure DevOps Pipeline is updated, reports will be updated when PRs are merged into `v2` branch and, eventually, `master`.


pjlittle:
Could add an ARM template or Terraform script to setup Azure Blob + static Website, but it's an easy, one-time setup.  Going to defer for now.

scripts/push-report.sh
```diff
@@ -0,0 +1,16 @@
...
+set -euo pipefail
+
+# This script appends code complexity reports over time. Given the amount of files, report files
+# are now stored on Azure Blob Storage. It's the source of truth - we download the current report,
+# run complexity analysis again, then push everything back to blob.
+rm -rf report
+mkdir -p report
+
+#NOTE: be sure to set AZURE_STORAGE_ACCOUNT and AZURE_STORAGE_KEY environment variables
+az storage blob download-batch -d report -s '$web'
```
mydiemho: where is `$web` set?

pjlittle: It's not a variable, which is why I'm using single quotes... when you enable static web site on Azure blob storage, it creates a container called `$web` by default. That's the container used to host all the web files. 

Not sure why they elected to use `$` to indicate that it's 'special' :man_shrugging:  If you're not careful, can definitely cause some problems in command line environments. The offering isn't flexible here; I can't find a way to specify another container name.

PR #575, opened 2019-02-14
Only hook video state change on non-preview assets AB#17076

tmiller-msft:
We're currently running unnecessary code for every video in the side bar by hooking the video state change callback and going through all the logic there. Only hook this callback if autoPlay is true.

### DIFF
--- a/src/react/components/common/assetPreview/videoAsset.test.tsx
+++ b/src/react/components/common/assetPreview/videoAsset.test.tsx
@@ -55,6 +55,18 @@ describe("Video Asset Component", () => {
         expect(wrapper.state().loaded).toBe(false);
     });
 
+    it("does not subscribe to state change callback when autoPlay is false", () => {
+        const testProps: IVideoAssetProps = { ...defaultProps,
+            autoPlay: false,
+        };
+        wrapper = createComponent(testProps);
+        mockLoaded();
+
+        const newAsset = MockFactory.createVideoTestAsset("new-video-asset");
+        wrapper.setProps({ asset: newAsset });
+        expect(videoPlayerMock.prototype.subscribeToStateChange).not.toBeCalled();
+    });
+
     it("seeks the video player to the spcified timestamp when timestamp changes", () => {
         const expectedTime = 10.2;
         wrapper = createComponent();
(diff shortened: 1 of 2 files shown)


src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -117,7 +117,12 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
     }
 
     public componentDidMount() {
-        this.videoPlayer.current.subscribeToStateChange(this.onVideoStateChange);
+        if (this.props.autoPlay) {
```
wbreza: Can you add a unit test to validate this behavior?

tmiller-msft: Done

src/react/components/common/assetPreview/videoAsset.test.tsx
```diff
@@ -55,6 +55,18 @@ describe("Video Asset Component", () => {
         expect(wrapper.state().loaded).toBe(false);
     });
 
+    it("does not subscribe to state change callback when autoPlay is false", () => {
```
mydiemho: shouldn't we another test for when autoPlay is true, that "subscribe" is called?

src/react/components/common/assetPreview/videoAsset.test.tsx
```diff
@@ -55,6 +55,18 @@ describe("Video Asset Component", () => {
...
+        const testProps: IVideoAssetProps = { ...defaultProps,
+            autoPlay: false,
+        };
+        wrapper = createComponent(testProps);
+        mockLoaded();
+
+        const newAsset = MockFactory.createVideoTestAsset("new-video-asset");
```
mydiemho: why create another asset when defaultProps already include a video asset?

src/react/components/common/assetPreview/videoAsset.test.tsx
tmiller-msft: Almost every other test would fail were this not the case. The change (and this test) were to handle the explicit case, adding another test to handle the default case (which is covered numerous times in the other tests) seems to be unnecessary.

src/react/components/common/assetPreview/videoAsset.test.tsx
tmiller-msft: To be explicit and clear what the test was doing, so people wouldn't have to go look up what default props was to figure out where this magic asset came from, along with ensuring a second asset was created.

PR #576, opened 2019-02-14
Fixes previous/next buttons to navigate between assets

wbreza:
- Fixes bug where editor page was attempting to set the selectedAsset in the sidebar to a non-root asset
- Renamed `parentAsset` to `rootAsset` to make this more clear in event handlers

### DIFF
--- a/src/react/components/pages/editorPage/editorPage.test.tsx
+++ b/src/react/components/pages/editorPage/editorPage.test.tsx
@@ -234,8 +234,8 @@ describe("Editor Page Component", () => {
             regions: editedImageAsset.regions,
         });
 
-        const matchingParentAsset = editorPage.state().assets.find((asset) => asset.id === imageAsset.id);
-        expect(matchingParentAsset.state).toEqual(AssetState.Tagged);
+        const matchingRootAsset = editorPage.state().assets.find((asset) => asset.id === imageAsset.id);
+        expect(matchingRootAsset.state).toEqual(AssetState.Tagged);
     });
 
     describe("Editing Video Assets", () => {
@@ -306,8 +306,8 @@ describe("Editor Page Component", () => {
             // Child asset is updated
             expect(saveMock.mock.calls[1][0]).toEqual(editedVideoFrame);
 
-            const matchingParentAsset = editorPage.state().assets.find((asset) => asset.id === videoAsset.id);
-            expect(matchingParentAsset.state).toEqual(AssetState.Tagged);
+            const matchingRootAsset = editorPage.state().assets.find((asset) => asset.id === videoAsset.id);
+            expect(matchingRootAsset.state).toEqual(AssetState.Tagged);
         });
     });
     describe("Basic toolbar test", () => {
--- a/src/react/components/pages/editorPage/editorPage.test.tsx
+++ b/src/react/components/pages/editorPage/editorPage.test.tsx
@@ -202,21 +202,65 @@ describe("Editor Page Component", () => {
         expect(saveProjectSpy).toBeCalledWith(expect.objectContaining(partialProject));
     });
 
+    it("When an image is updated the asset metadata is updated", async () => {
+        const testProject = MockFactory.createTestProject("TestProject");
+        const store = createStore(testProject, true);
+        const props = MockFactory.editorPageProps(testProject.id);
+        const wrapper = createComponent(store, props);
+        const imageAsset = testAssets[0];
+
+        await MockFactory.flushUi();
+        wrapper.update();
+
+        const editedImageAsset: IAssetMetadata = {
+            asset: imageAsset,
+            regions: [MockFactory.createMockRegion()],
+        };
+
+        const saveMock = assetServiceMock.prototype.save as jest.Mock;
+        saveMock.mockClear();
+
+        wrapper.find(Canvas).props().onAssetMetadataChanged(editedImageAsset);
+        await MockFactory.flushUi();
+
+        const editorPage = wrapper.find(EditorPage).childAt(0) as ReactWrapper<IEditorPageProps, IEditorPageState>;
+
+        // Image asset is updated
+        expect(assetServiceMock.prototype.save).toBeCalledWith({
+            asset: {
+                ...imageAsset,
+                state: AssetState.Tagged,
+            },
+            regions: editedImageAsset.regions,
+        });
+
+        const matchingParentAsset = editorPage.state().assets.find((asset) => asset.id === imageAsset.id);
+        expect(matchingParentAsset.state).toEqual(AssetState.Tagged);
+    });
+
     describe("Editing Video Assets", () => {
-        it("Child assets are not included within editor page state", async () => {
+        let wrapper: ReactWrapper;
+        let videoAsset: IAsset;
+        let videoFrames: IAsset[];
+
+        beforeEach(async () => {
             const testProject = MockFactory.createTestProject("TestProject");
-            const videoAsset = MockFactory.createVideoTestAsset("TestVideo");
-            const videoFrames = MockFactory.createChildVideoAssets(videoAsset);
+            videoAsset = MockFactory.createVideoTestAsset("TestVideo");
+            videoFrames = MockFactory.createChildVideoAssets(videoAsset);
             const projectAssets = [videoAsset].concat(videoFrames);
             testProject.assets = _.keyBy(projectAssets, (asset) => asset.id);
 
             const store = createStore(testProject, true);
             const props = MockFactory.editorPageProps(testProject.id);
 
-            const wrapper = createComponent(store, props);
-            const editorPage = wrapper.find(EditorPage).childAt(0) as ReactWrapper<IEditorPageProps, IEditorPageState>;
+            wrapper = createComponent(store, props);
 
             await MockFactory.flushUi();
+            wrapper.update();
+        });
+
+        it("Child assets are not included within editor page state", () => {
+            const editorPage = wrapper.find(EditorPage).childAt(0) as ReactWrapper<IEditorPageProps, IEditorPageState>;
 
             expect(editorPage.state().assets.length).toEqual(testAssets.length + 1);
             expect(editorPage.state().selectedAsset.asset).toEqual({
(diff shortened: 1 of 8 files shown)


src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -185,6 +184,30 @@ describe("Editor Page Component", () => {
...
+            const store = createStore(testProject, true);
+            const props = MockFactory.editorPageProps(testProject.id);
+
+            const wrapper = createComponent(store, props);
+            const editorPage = wrapper.find(EditorPage).childAt(0) as ReactWrapper<IEditorPageProps, IEditorPageState>;
+
+            await MockFactory.flushUi();
+
+            expect(editorPage.state().assets.length).toEqual(testAssets.length + 1);
+            expect(editorPage.state().selectedAsset.asset).toEqual({
```
mydiemho: `selectedAsset` should be rename (it was confusing to see what I thought is an asset object containing an asset property)

src/services/assetService.ts
```diff
@@ -151,11 +151,14 @@ export class AssetService {
...
 
+        const fileName = `${metadata.asset.id}${constants.assetMetadataFileExtension}`;
+
         // Only save asset metadata if asset is in a tagged state
         // Otherwise primary asset information is already persisted in the project file.
         if (metadata.asset.state === AssetState.Tagged) {
-            const fileName = `${metadata.asset.id}${constants.assetMetadataFileExtension}`;
             await this.storageProvider.writeText(fileName, JSON.stringify(metadata, null, 4));
+        } else {
+            await this.storageProvider.deleteFile(fileName);
```
mydiemho: why are we deleting file here?

src/react/components/pages/editorPage/editorSideBar.tsx
```diff
@@ -21,6 +21,7 @@ export interface IEditorSideBarProps {
  */
 export interface IEditorSideBarState {
     selectedAsset: IAsset;
+    scrollToIndex: number;
```
mydiemho: I would suggest pulling this change to a different PR since it's fixing a different issues

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -229,20 +229,51 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         }
     }
 
+    private isChildAssetType = (asset: IAsset): boolean => {
```
mydiemho: the naming of this method is confusing.  As discussed, we should call it `isContainerType` or something

src/react/components/pages/editorPage/editorPage.test.tsx
wbreza: The `selectedAsset` on the state is the selected asset that is currently loaded in the editor.

src/react/components/pages/editorPage/editorPage.tsx
wbreza: Renamed to `isTaggableAssetType`.

src/services/assetService.ts
wbreza: If the asset is not in a tagged state then there is no reason to keep it around since it doesn't contain and region data.

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -203,6 +202,114 @@ describe("Editor Page Component", () => {
...
+            regions: [MockFactory.createMockRegion()],
+        };
+
+        const saveMock = assetServiceMock.prototype.save as jest.Mock;
+        saveMock.mockClear();
+
+        wrapper.find(Canvas).props().onAssetMetadataChanged(editedImageAsset);
+        await MockFactory.flushUi();
+
+        const editorPage = wrapper.find(EditorPage).childAt(0) as ReactWrapper<IEditorPageProps, IEditorPageState>;
```
mydiemho: what's the difference between your version and doing

```
const editorPage: ReactWrapper<IEditorPageProps, IEditorPageState> = wrapper.find(EditorPage).childAt(0)
```

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -203,6 +202,114 @@ describe("Editor Page Component", () => {
...
+        await MockFactory.flushUi();
+        wrapper.update();
+
+        const editedImageAsset: IAssetMetadata = {
+            asset: imageAsset,
+            regions: [MockFactory.createMockRegion()],
+        };
+
+        const saveMock = assetServiceMock.prototype.save as jest.Mock;
+        saveMock.mockClear();
```
mydiemho: why do we have to call `mockClear` when the mock is brand new?

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -203,6 +202,114 @@ describe("Editor Page Component", () => {
...
+            const store = createStore(testProject, true);
+            const props = MockFactory.editorPageProps(testProject.id);
+
+            wrapper = createComponent(store, props);
+
+            await MockFactory.flushUi();
+            wrapper.update();
+        });
+
+        it("Child assets are not included within editor page state", () => {
```
mydiemho: does child here refer to the assets inside a container asset (so image wouldn't count)?

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -203,6 +202,114 @@ describe("Editor Page Component", () => {
...
+
+            const expectedParentVideoMetadata: IAssetMetadata = {
+                asset: {
+                    ...videoAsset,
+                    state: AssetState.Tagged,
+                },
+                regions: [],
+            };
+
+            // Called 2 times, one for parent and once for child.
```
mydiemho: Nit: one/once

src/react/components/pages/editorPage/editorPage.test.tsx
wbreza: The mock isn't brand new, its just being casted to a proper mock type after being instantiated up above in the before Each.

wbreza: And we want to reset the mock call count since by default the editor page selects the first asset in the list.

src/react/components/pages/editorPage/editorPage.test.tsx
wbreza: Compiler doesn't the type of `.childAt(0)` so you have to cast it... i believe the way you mentioned would cause a TS compile error.  Will verify

src/react/components/pages/editorPage/editorPage.test.tsx
wbreza: Child meaning if the asset has a parent, ex) a VideoFrame

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -239,20 +239,53 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+        return asset.type === AssetType.Image || asset.type === AssetType.VideoFrame;
+    }
+
     /**
      * Raised when the selected asset has been changed.
      * This can either be a parent or child asset
      */
     private onAssetMetadataChanged = async (assetMetadata: IAssetMetadata): Promise<void> => {
-        assetMetadata.asset.state = assetMetadata.regions.length > 0 ? AssetState.Tagged : AssetState.Visited;
+        const parentAsset = { ...(assetMetadata.asset.parent || assetMetadata.asset) };
```
mydiemho: Nit: we're calling this `parentAsset` but value could be nonparent asset

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -239,20 +239,53 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+
+        if (this.isTaggableAssetType(assetMetadata.asset)) {
+            assetMetadata.asset.state = assetMetadata.regions.length > 0 ? AssetState.Tagged : AssetState.Visited;
+        } else if (assetMetadata.asset.state === AssetState.NotVisited) {
+            assetMetadata.asset.state = AssetState.Visited;
+        }
+
+        // Update parent asset if not already in the "Tagged" state
+        if (parentAsset.id === assetMetadata.asset.id) {
```
mydiemho: this confused me.  we get `parentAsset` from `assetMetadata` so wouldn't id be the same?

src/react/components/pages/editorPage/editorPage.tsx
wbreza: Renamed to `rootAsset` to make this more clearn.

src/react/components/pages/editorPage/editorPage.tsx
wbreza: Renamed to `rootAsset` to make this more clear.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -332,18 +386,24 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
 
         this.loadingProjectAssets = true;
 
-        const projectAssets = _.values(this.props.project.assets);
+        // Get all root project assets
+        const rootProjectAssets = _.values(this.props.project.assets)
```
mydiemho: I would rename this `projectRootAssets` since it's looking for all the root assets in a project

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -239,20 +239,63 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+     */
+    private isTaggableAssetType = (asset: IAsset): boolean => {
+        return asset.type === AssetType.Image || asset.type === AssetType.VideoFrame;
+    }
+
     /**
      * Raised when the selected asset has been changed.
      * This can either be a parent or child asset
      */
     private onAssetMetadataChanged = async (assetMetadata: IAssetMetadata): Promise<void> => {
```
mydiemho: this method is still really confusing to me :(

tbarlow12: @mydiemho the purpose of it or the actual implementation?

wbreza: @mydiemho - Let's discuss offline.

PR #577, opened 2019-02-14
EditorPage does not have correct state when navigating directly via URL- AB#17079

tmiller-msft:
When navigating directly to a project edit page via URL, the initial state is invalid, and is not updated when the correct data is retrieved, and this fixes that.

### DIFF
--- a/src/react/components/pages/editorPage/editorPage.tsx
+++ b/src/react/components/pages/editorPage/editorPage.tsx
@@ -102,6 +102,16 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         if (this.props.project && this.state.assets.length === 0) {
             await this.loadProjectAssets();
         }
+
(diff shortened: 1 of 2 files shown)


src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -102,6 +102,16 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         if (this.props.project && this.state.assets.length === 0) {
             await this.loadProjectAssets();
         }
+
+        // Navigating directly to the page via URL (ie, http://vott/projects/a1b2c3dEf/edit) sets the default state
+        // before props has been set, this updates the project and additional settings to be valid once props are
+        // retrieved.
+        if (!this.state.project && this.props.project) {
+            this.setState({
+                project: this.props.project,
```
wbreza: `this.state.project` is not actually used in the editor page.  I believe this can be deprecated.

PR #578, opened 2019-02-14
Fix Asset detection from path

JacopoMangiavacchi:
Resolve issue detecting asset type in case the file name contain more dot separator:

- ie file.jpg.tfrecord

### DIFF
--- a/src/services/assetService.ts
+++ b/src/services/assetService.ts
@@ -33,9 +33,13 @@ export class AssetService {
         // fileNameParts[0] = "video"
         // fileNameParts[1] = "mp4"
         // fileNameParts[2] = "t=5"
-        const fileNameParts = pathParts[pathParts.length - 1].split(/[\.\?#]/);
-        fileName = fileName || `${fileNameParts[0]}.${fileNameParts[1]}`;
-        const assetFormat = fileNameParts.length >= 2 ? fileNameParts[1] : "";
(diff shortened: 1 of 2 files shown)


src/services/assetService.ts
```diff
@@ -33,9 +33,13 @@ export class AssetService {
         // fileNameParts[0] = "video"
         // fileNameParts[1] = "mp4"
         // fileNameParts[2] = "t=5"
-        const fileNameParts = pathParts[pathParts.length - 1].split(/[\.\?#]/);
-        fileName = fileName || `${fileNameParts[0]}.${fileNameParts[1]}`;
-        const assetFormat = fileNameParts.length >= 2 ? fileNameParts[1] : "";
+        fileName = pathParts[pathParts.length - 1];
```
wbreza: You are overriding the `filename` argument in all cases now... If a filename is specified we should automatically use that one...

src/services/assetService.ts
```diff
@@ -33,9 +33,13 @@ export class AssetService {
...
+        const fileNameParts = fileName.split(/[\.\?#]/);
+        let assetFormat = fileNameParts[fileNameParts.length - 1];
+        if (fileNameParts.length >= 2 && this.getAssetType(fileNameParts[1]) === AssetType.Video) {
```
wbreza: We should not need to have any special case handling for any specific asset type

PR #581, opened 2019-02-14
Updating report CI scripts to be more robust.

pjlittle:
Complexity CI is _finally_ working end-to-end, retaining history for each merged PR.

- Updated scripts to be more explicit with paths, passing them via command line args
- Added more logging
- Updated docs

Was able to test this e2e by pointing Azure DevOps to my branch.

### DIFF
--- a/scripts/complexity-analysis.js
+++ b/scripts/complexity-analysis.js
@@ -4,7 +4,9 @@ const plato = require('es6-plato');
 // parse command line args
 const optionDefinitions = [
     { name: 'src', type: String, defaultValue: './es6-src/**' },
-    { name: 'output', type: String, defaultValue: './report' }
+    { name: 'output', type: String, defaultValue: './report' },
+    { name: 'version', type: String, defaultValue: '2.0.0' },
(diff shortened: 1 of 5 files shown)


scripts/complexity-analysis.js
```diff
@@ -4,7 +4,9 @@ const plato = require('es6-plato');
 // parse command line args
 const optionDefinitions = [
     { name: 'src', type: String, defaultValue: './es6-src/**' },
-    { name: 'output', type: String, defaultValue: './report' }
+    { name: 'output', type: String, defaultValue: './report' },
+    { name: 'version', type: String, defaultValue: '2.0.0' },
```
wbreza: What is this `2.0.0` here? Is that VoTT version?  If so can we pull out of package.json like you do above?

wbreza: Or is the the value only if version is not already supplied?

pjlittle: just a default value, if one isn't passed in.

PR #583, opened 2019-02-15
Elhalper/import service

elizabethhalper:
Import service converts a v1 project to v2 format


src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -138,17 +138,17 @@ describe("Editor Canvas", () => {
         expect(wrapper.state().selectedRegions).toEqual([assetMetadata.regions[assetMetadata.regions.length - 1]]);
     });
 
-    it("onRegionMove edits region info in asset", () => {
-        const canvas = wrapper.instance();
-        const testRegion = MockFactory.createTestRegion("test-region");
+    // it("onRegionMove edits region info in asset", () => {
```
wbreza: Any reason why this test is commented out?

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -48,7 +48,7 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
         const sz = document.getElementById("editor-zone") as HTMLDivElement;
         this.editor = new CanvasTools.Editor(sz);
         this.editor.onSelectionEnd = this.onSelectionEnd;
-        this.editor.onRegionMoveEnd = this.onRegionMoveEnd;
+        // this.editor.onRegionMoveEnd = this.onRegionMoveEnd;
```
wbreza: Why is this commented out?

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -150,4 +189,16 @@ export default class HomePage extends React.Component<IHomepageProps> {
             throw new AppError(ErrorCode.ProjectDeleteError, "Error deleting project file");
         }
     }
+
+    private convertProject = async (project: IFileInfo) => {
+        const importService = new ImportService();
+        let projectJson;
+        try {
+            projectJson = await importService.convertV1(project);
```
wbreza: Instead of calling out `v1` in the method name I would just call it convert and leave it up the the import service to do the required transformations / version detection.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -139,8 +170,16 @@ export default class HomePage extends React.Component<IHomepageProps> {
     }
 
     private loadSelectedProject = async (project: IProject) => {
-        await this.props.actions.loadProject(project);
-        this.props.history.push(`/projects/${project.id}/edit`);
+        console.log(project);
```
wbreza: I don't think that this method should require changing... any project passed into this method should already be a valid v2 project and should open the same as any other project.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -150,4 +189,16 @@ export default class HomePage extends React.Component<IHomepageProps> {
             throw new AppError(ErrorCode.ProjectDeleteError, "Error deleting project file");
         }
     }
+
+    private convertProject = async (project: IFileInfo) => {
```
wbreza: I recommend renaming many of this variables:
`projects` on the input params is not a project, but rather the v1Project file info.
`projectJson` looks to be the converted project in v2 format rather than a JSON string.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -150,4 +189,16 @@ export default class HomePage extends React.Component<IHomepageProps> {
...
+        const importService = new ImportService();
+        let projectJson;
+        try {
+            projectJson = await importService.convertV1(project);
+        } catch (e) {
+            throw new AppError(ErrorCode.ProjectUploadError, "Error converting v1 project file");
```
wbreza: Looks like this should throw the convert error code vs. project upload error.

src/redux/actions/applicationActions.test.ts
```diff
@@ -61,4 +69,36 @@ describe("Application Redux Actions", () => {
...
+        const actions = store.getActions();
+
+        expect(actions.length).toEqual(1);
+        expect(actions[0]).toEqual({
+            type: ActionTypes.ENSURE_SECURITY_TOKEN_SUCCESS,
+            payload: testProject,
+        });
+
+        expect(result).toEqual(appSettings);
+        expect(testProject.securityToken).toEqual("Project TestProject Token");
```
wbreza: I would also confirm that the security token was correctly added into the appSettings.securityTokens array.

src/redux/actions/applicationActions.ts
```diff
@@ -50,6 +53,39 @@ export function saveAppSettings(appSettings: IAppSettings): (dispath: Dispatch)
...
+ * @param project The project to validate
+ */
+export function ensureSecurityToken(project: IProject):
+    (dispatch: Dispatch, getState: () => IApplicationState) => Promise<IAppSettings> {
+    return async (dispatch: Dispatch, getState: () => IApplicationState) => {
+        const appState = getState();
+        let securityToken = appState.appSettings.securityTokens
+            .find((st) => st.name === project.securityToken);
+
+        if (securityToken) {
```
wbreza: Since you have already declared your function as `async` you do not need to wrap all your returns in a promise... it will automatically be wrapped by the compiler...

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+        generatedAssetMetadata = this.generateAssets(project, assetService);
+
+        const saveAssets = generatedAssetMetadata.map((assetMetadata) => {
+            assetMetadata.then((metadata) => {
+                return assetService.save(metadata);
+            });
+        });
+
+        try {
+            Promise.all(saveAssets);
```
wbreza: You need to `await` this call to block before returning the converted project.

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+
+        parsedTags = this.parseTags(originalProject);
+
+        connections = this.generateConnections(project);
+
+        // map v1 values to v2 values
+        convertedProject = {
+            id: shortid.generate(),
+            name: project.file.name.split(".")[0],
+            version: "v1-to-v2",
```
wbreza: As mentioned previously let's pull in the actual app version from the `package.json`

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+            securityToken: `${project.file.name.split(".")[0]} Token`,
+            description: "Converted V1 Project",
+            tags: parsedTags,
+            sourceConnection: connections[0],
```
wbreza: By default we should reuse the same connection for both the source & target.

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        return convertedProject;
+    }
+
+    /**
+     * Generate connections from v1 project file location
+     * @param project - V1 Project Content and File Information
+     */
+    private generateConnections(project: any): IConnection[] {
```
wbreza: As mentioned above, let's reuse the same connection for both the source and target connections... they can always go back and change if needed.

wbreza: Please avoid use of `any` type in `project` input argument.

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+            finalTags.push(newTag);
+        }
+        return finalTags;
+    }
+
+    /**
+     * Generate assets based on V1 Project frames and regions
+     * @param project - V1 Project Content and File Information
+     */
+    private generateAssets(project: any, assetService: AssetService): Array<Promise<IAssetMetadata>> {
```
wbreza: Avoid use of `any` in `project` input argument.

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+        let originalProject: IV1Project;
+        const generatedAssetMetadata: Array<Promise<IAssetMetadata>> = [];
+        let generatedRegion: IRegion;
+        let assetState: AssetState;
+        const currentTagColorIndex = randomIntInRange(0, TagColors.length);
+
+        originalProject = JSON.parse(project.content);
+
+        for (const frameName in originalProject.frames) {
```
wbreza: This loop is rather complex with lots of inner loops going on... Let's refactor to split some of this log out into smaller bitsize methods that you can then reuse here.

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+            if (originalProject.frames.hasOwnProperty(frameName)) {
+                const frameRegions = originalProject.frames[frameName];
+                const asset = AssetService.createAssetFromFilePath(
+                    `file:${project.file.path.replace(/[^\/]*$/, "")}${frameName}`);
+                const assetMetadata = assetService.getAssetMetadata(asset).then((metadata) => {
```
wbreza: Let's standardize on `async` / `await` vs. promise `then` callbacks where ever possible.

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+                    // assetMetadata = metadata;
+                    assetState = originalProject.visitedFrames.indexOf(frameName) > -1 && frameRegions.length > 0
+                                ? AssetState.Tagged : (originalProject.visitedFrames.indexOf(frameName) > -1
+                                ? AssetState.Visited : AssetState.NotVisited);
+                    metadata.asset.state = assetState;
+
+                    /*
```
wbreza: Removed unneeded commented out code.

src/services/importService.ts
```diff
@@ -0,0 +1,191 @@
...
+                                newTag = {
+                                    name: tag,
+                                    color: TagColors[(currentTagColorIndex + 1) % TagColors.length],
+                                };
+                                return newTag;
+                            });
+                    */
+
+                    for (const region of frameRegions) {
+                        generatedRegion = {
```
wbreza: Looks like `generatedRegion` should be scoped to within this loop vs. being scoped outside of this loop - otherwise you are redefining the region on every iteration.

src/services/importService.ts
wbreza: I would update this method to return the final asset metadata for the assets vs. the promises.

src/services/projectService.ts
```diff
@@ -53,6 +54,9 @@ export default class ProjectService implements IProjectService {
                     project.id = shortid.generate();
                 }
 
+                // automate this eventually
+                project.version = "2.0.0-preview.1";
```
wbreza: https://stackoverflow.com/questions/45978230/get-version-number-from-package-json-in-react-redux-create-react-app

src/services/importService.ts
wbreza: I would break this down into many smaller methods that convert v1Type to v2Type.  Then this function should essentially becoming a simple orchestrator to convert the project.

wbreza:
@elizabethhalper How is this PR different from your WIP one? Let's close the one that is not in use.

PR #584, opened 2019-02-15
Fixes [AB#17029] fix sidebar so it's in sync with asset navigation

mydiemho:
Sidebar is not updated when moving between asset with tool bar button or keyboard button.  Update to always focus on selected asset on first asset on first load.

### DIFF
--- a/src/react/components/pages/editorPage/editorSideBar.test.tsx
+++ b/src/react/components/pages/editorPage/editorSideBar.test.tsx
@@ -5,7 +5,7 @@ import { AutoSizer, List } from "react-virtualized";
 import MockFactory from "../../../../common/mockFactory";
 
 describe("Editor SideBar", () => {
-    const onSelectAssetHanlder = jest.fn();
+    const onSelectAssetHandler = jest.fn();
     const testAssets = MockFactory.createTestAssets();
 
     function createComponent(props: IEditorSideBarProps): ReactWrapper<IEditorSideBarProps, IEditorSideBarState> {
@@ -15,7 +15,7 @@ describe("Editor SideBar", () => {
     it("Component renders correctly", () => {
         const props: IEditorSideBarProps = {
             assets: testAssets,
-            onAssetSelected: onSelectAssetHanlder,
+            onAssetSelected: onSelectAssetHandler,
         };
 
         const wrapper = createComponent(props);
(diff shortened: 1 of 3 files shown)


src/react/components/pages/editorPage/editorSideBar.test.tsx
```diff
@@ -27,89 +27,82 @@ describe("Editor SideBar", () => {
...
-            expect(wrapper.state().selectedAsset).toEqual(testAssets[0]);
+            const state = wrapper.state();
+            expect(state.selectedAsset).toEqual(testAssets[selectedAssetIndex]);
+            expect(state.scrollToIndex).toEqual(selectedAssetIndex);
+
             done();
         });
     });
 
-    it("Calls onAssetSelected handler when an asset is selected", (done) => {
```
mydiemho: removed this test because there isn't a way to access the asset element inside a react-virtualized list.  Without this element, we cannot simulate a click, therefore cannot test whether onAssetSelected is called.

src/react/components/pages/editorPage/editorSideBar.tsx
```diff
@@ -56,6 +63,8 @@ export default class EditorSideBar extends React.Component<IEditorSideBarProps,
                             rowHeight={233}
                             rowRenderer={this.rowRenderer}
                             overscanRowCount={2}
+                            scrollToAlignment="center"
```
mydiemho: align the selected asset to middle of the list

src/react/components/pages/editorPage/editorSideBar.test.tsx
```diff
@@ -5,7 +5,7 @@ import { AutoSizer, List } from "react-virtualized";
 import MockFactory from "../../../../common/mockFactory";
 
 describe("Editor SideBar", () => {
-    const onSelectAssetHanlder = jest.fn();
+    const onSelectAssetHandler = jest.fn();
```
wbreza: Thanks for fixing spelling 👍

src/react/components/pages/editorPage/editorSideBar.tsx
wbreza: What other options does it support?  Maybe "top" would be better if that is an option?

mydiemho: It also support "start" and "end".  Honestly, I prefer center since I tend to focus on the center.

https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md

JacopoMangiavacchi: Agree probably "center" is best option in our scenario.

Generally speaking the scrolling after a GOTO on listview is "start" or "end" depending on where you come from (previous selection).  If you want to easily try this open powerpoint or preview on macOS and try find or goto different page on preview.

PR #585, opened 2019-02-15
feat: Export video frames AB#16583

PIC123:
Adds ability to export videos, treating video frames as images

### DIFF
--- a/src/common/extensions/array.ts
+++ b/src/common/extensions/array.ts
@@ -1,4 +1,4 @@
-import Guard from "./guard";
+import Guard from "../guard";
 
 /**
  * Processes items in the array within the specified batch size (default: 5)
@@ -6,12 +6,13 @@ import Guard from "./guard";
  * @param action The action to perform on each item in the array
  * @param batchSize The batch size for actions to perform in parallel (default: 5)
  */
-export async function arrayForEachAsync<T>(
+export async function forEachAsync<T>(
     this: T[],
     action: (item: T) => Promise<void>,
     batchSize: number = 5): Promise<void> {
     Guard.null(this);
     Guard.null(action);
+    Guard.expression(batchSize, (value) => value > 0);
 
     const all: T[] = [...this];
 
@@ -33,12 +34,13 @@ export async function arrayForEachAsync<T>(
  * @param action The transformer action to perform on each item in the array
  * @param batchSize The batch size for actions to perform in parallel (default: 5);
  */
-export async function arrayMapAsync<T, R>(
+export async function mapAsync<T, R>(
     this: T[],
     action: (item: T) => Promise<R>,
     batchSize: number = 5): Promise<R[]> {
     Guard.null(this);
     Guard.null(action);
+    Guard.expression(batchSize, (value) => value > 0);
 
     let results: R[] = [];
     const all: T[] = [...this];
--- a/src/common/extensions/map.ts
+++ b/src/common/extensions/map.ts
@@ -1,4 +1,4 @@
-import Guard from "./guard";
+import Guard from "../guard";
 
 /**
  * Processes items in the map within the specified batch size (default: 5)
@@ -6,12 +6,13 @@ import Guard from "./guard";
  * @param action The action to perform on each item in the map
  * @param batchSize The batch size for actions to perform in parallel (default: 5)
  */
-export async function mapForEachAsync<K, V>(
+export async function forEachAsync<K, V>(
     this: Map<K, V>,
     action: (value: V, key: K) => Promise<void>,
     batchSize: number = 5): Promise<void> {
     Guard.null(this);
     Guard.null(action);
+    Guard.expression(batchSize, (value) => value > 0);
 
     const all: Array<[K, V]> = [...this.entries()];
 
--- a/src/common/htmlFileReader.test.ts
+++ b/src/common/htmlFileReader.test.ts
@@ -94,28 +94,6 @@ describe("Html File Reader", () => {
             expect(axios.get).toBeCalledWith(asset.path, { responseType: "blob" });
         });
 
-        it("Downloads a blob from the asset path", async () => {
-            const testAssetVideo = MockFactory.createVideoTestAsset("video-test");
-            const testAssetFrame = MockFactory.createChildVideoAsset(testAssetVideo, 0);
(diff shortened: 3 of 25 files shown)


src/common/htmlFileReader.ts
```diff
@@ -73,13 +75,42 @@ export default class HtmlFileReader {
...
+                const ctx = canvas.getContext("2d");
+                ctx.drawImage(video, 0, 0, canvas.width, canvas.height);
+                canvas.toBlob((blob) => {
+                    resolve(blob);
+                });
+            };
+            video.onerror = (e) => {
+                reject(e);
+            };
+            video.src = asset.path;
```
JacopoMangiavacchi: Even if the video object is created only on the virtual dom it seams to me that loading anytime the entire video files in memory for each individual frame we want to export could be an issue on both performance and memory consumption.  Just wonder if it is not better to implement a batch interface, load the video file once and return a list of all requested frames.

PIC123: So this is for the first pass still. I was hoping to see if the app was smart enough to do any caching of the data. If there are performance issue, which is definitely a possibility, it will be a batch process that just loads the video once.

src/common/htmlFileReader.ts
```diff
@@ -73,13 +75,28 @@ export default class HtmlFileReader {
...
+                throw new Error("Error downloading asset binary");
+            }
+            data = await response.data;
         }
 
-        return response.data;
+        return data;
+    }
+
+    public static dataURItoBlob(dataURI) {
```
wbreza: Is this being used anywhere?

src/providers/export/exportProvider.ts
```diff
@@ -52,13 +57,41 @@ export interface IExportProvider {
...
 
     public abstract export(): Promise<void> | Promise<IExportResults>;
 
+    /**
+     * Gets the assets that are configured to be exported based on the configured asset state
+     */
+    public async getAssetsForExport(): Promise<IAssetMetadata[]> {
+        let predicate: (asset: IAsset) => boolean = null;
+
+        // @ts-ignore
```
wbreza: What are we ignoring?  Doesn't look like this should be here.

src/common/htmlFileReader.ts
```diff
@@ -92,6 +109,40 @@ export default class HtmlFileReader {
         return new Uint8Array(byteArray);
     }
 
+    public static async getAssetFrameImage(asset: IAsset) {
```
wbreza: I don't see any unit tests to verify this video frame extraction

src/common/htmlFileReader.test.ts
```diff
@@ -93,6 +94,19 @@ describe("Html File Reader", () => {
...
+            const testAssetVideo = MockFactory.createVideoTestAsset("video-test");
+            const testAssetFrame = MockFactory.createChildVideoAsset(testAssetVideo, 0);
+            HtmlFileReader.getAssetFrameImage = jest.fn((asset) => {
+                return Promise.resolve(new Blob(["Some binary data"]));
+            });
+
+            const result = await HtmlFileReader.getAssetBlob(testAssetFrame);
+            expect(result).not.toBeNull();
+            expect(result).toBeInstanceOf(Blob);
+            expect(HtmlFileReader.getAssetFrameImage).toBeCalledWith(testAssetFrame);
```
wbreza: This test is not actually testing the logic of your `getAssetFrameImage` method since you are mocking it.  Lets write some tests to cover this better.

src/common/extensions/array.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export async function forEachAsync<T>(
+    this: T[],
+    action: (item: T) => Promise<void>,
+    batchSize: number = 5): Promise<void> {
+    Guard.null(this);
+    Guard.null(action);
+    Guard.expression(batchSize, (value) => value > 0);
+
+    const all: T[] = [...this];
```
JacopoMangiavacchi: As this is a generic extension for any array and potentially it could be reused in other cases with different requirements (i.e. array and/or element size) I just wonder why we copy the original array on a mutable variable instead of directly using the "this" array with a index instead of pop() from this copy.

src/common/extensions/array.ts
```diff
@@ -0,0 +1,61 @@
...
+export async function mapAsync<T, R>(
+    this: T[],
+    action: (item: T) => Promise<R>,
+    batchSize: number = 5): Promise<R[]> {
+    Guard.null(this);
+    Guard.null(action);
+    Guard.expression(batchSize, (value) => value > 0);
+
+    let results: R[] = [];
+    const all: T[] = [...this];
```
JacopoMangiavacchi: Same comment as above

src/common/extensions/map.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export async function forEachAsync<K, V>(
+    this: Map<K, V>,
+    action: (value: V, key: K) => Promise<void>,
+    batchSize: number = 5): Promise<void> {
+    Guard.null(this);
+    Guard.null(action);
+    Guard.expression(batchSize, (value) => value > 0);
+
+    const all: Array<[K, V]> = [...this.entries()];
```
JacopoMangiavacchi: Again, same comment here

src/common/extensions/array.ts
wbreza: `pop()` from the original array would mutate the original array which would be an unintended side effect.

src/common/extensions/array.ts
wbreza: Same reply as above - we don't want to mutate the original array.

src/common/extensions/map.ts
wbreza: Same - We don't want to mutate the original map

src/common/extensions/array.ts
JacopoMangiavacchi: @wbreza of course.  Indeed I was suggesting to use an index iterator and do not mutate neither the original or the copy

src/common/extensions/array.ts
JacopoMangiavacchi: Replied above

src/common/extensions/map.ts
JacopoMangiavacchi: Replied above

PR #586, opened 2019-02-15
Add tag to project while importing TFRecords [AB#17001]

JacopoMangiavacchi:
- Import tag metadata from TFRecords file and save asset metadata with region/tag info and tagged state
- Add tags to Project while automatically import new tag information from the asset metadata
- Refresh Tag component on the EditorToolbar with the new founded tags
- Implement unit tests for testing tag detection, project update and tag component refresh

### DIFF
--- a/src/common/mockFactory.ts
+++ b/src/common/mockFactory.ts
@@ -133,7 +133,7 @@ export default class MockFactory {
     /**
      * Creates a mock region
      */
-    public static createMockRegion(id?: string): IRegion {
+    public static createMockRegion(id?: string, forcedNewTag?: string): IRegion {
         const mockTag: ITag = MockFactory.createTestTag();
 
         const mockStartPoint: IPoint = {
@@ -156,7 +156,7 @@ export default class MockFactory {
         const mockRegion: IRegion = {
             id: id || "id",
             type: RegionType.Rectangle,
-            tags: [mockTag.name],
+            tags: [forcedNewTag ? forcedNewTag : mockTag.name],
             points: [mockStartPoint, mockEndPoint],
             boundingBox: mockBoundingBox,
         };
--- a/src/common/mockFactory.ts
+++ b/src/common/mockFactory.ts
@@ -133,7 +133,7 @@ export default class MockFactory {
     /**
      * Creates a mock region
      */
-    public static createMockRegion(id?: string, forcedNewTag?: string): IRegion {
+    public static createMockRegion(id?: string): IRegion {
         const mockTag: ITag = MockFactory.createTestTag();
 
         const mockStartPoint: IPoint = {
@@ -156,7 +156,7 @@ export default class MockFactory {
         const mockRegion: IRegion = {
             id: id || "id",
             type: RegionType.Rectangle,
-            tags: [forcedNewTag ? forcedNewTag : mockTag.name],
+            tags: [mockTag.name],
             points: [mockStartPoint, mockEndPoint],
             boundingBox: mockBoundingBox,
         };
--- a/src/react/components/pages/editorPage/canvas.tsx
+++ b/src/react/components/pages/editorPage/canvas.tsx
@@ -293,14 +293,12 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState> 
      * Positions the canvas tools drawing surface to be exactly over the asset content
      */
     private positionCanvas = (contentSource: ContentSource) => {
-        if (contentSource) {
-            const canvas = this.canvasZone.current;
-            canvas.style.top = `${contentSource.offsetTop}px`;
-            canvas.style.left = `${contentSource.offsetLeft}px`;
-            canvas.style.width = `${contentSource.offsetWidth}px`;
-            canvas.style.height = `${contentSource.offsetHeight}px`;
-            this.editor.resize(contentSource.offsetWidth, contentSource.offsetHeight);
-        }
+        const canvas = this.canvasZone.current;
+        canvas.style.top = `${contentSource.offsetTop}px`;
+        canvas.style.left = `${contentSource.offsetLeft}px`;
+        canvas.style.width = `${contentSource.offsetWidth}px`;
+        canvas.style.height = `${contentSource.offsetHeight}px`;
+        this.editor.resize(contentSource.offsetWidth, contentSource.offsetHeight);
     }
 
     /**
--- a/src/react/components/pages/editorPage/canvas.tsx
+++ b/src/react/components/pages/editorPage/canvas.tsx
@@ -293,12 +293,14 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState> 
      * Positions the canvas tools drawing surface to be exactly over the asset content
      */
     private positionCanvas = (contentSource: ContentSource) => {
-        const canvas = this.canvasZone.current;
-        canvas.style.top = `${contentSource.offsetTop}px`;
-        canvas.style.left = `${contentSource.offsetLeft}px`;
-        canvas.style.width = `${contentSource.offsetWidth}px`;
-        canvas.style.height = `${contentSource.offsetHeight}px`;
-        this.editor.resize(contentSource.offsetWidth, contentSource.offsetHeight);
+        if (contentSource) {
+            const canvas = this.canvasZone.current;
+            canvas.style.top = `${contentSource.offsetTop}px`;
+            canvas.style.left = `${contentSource.offsetLeft}px`;
+            canvas.style.width = `${contentSource.offsetWidth}px`;
+            canvas.style.height = `${contentSource.offsetHeight}px`;
+            this.editor.resize(contentSource.offsetWidth, contentSource.offsetHeight);
+        }
     }
 
     /**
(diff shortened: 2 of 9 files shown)


src/react/components/pages/editorPage/editorFooter.tsx
```diff
@@ -37,6 +37,14 @@ export default class EditorFooter extends React.Component<IEditorFooterProps, IE
         this.onTagsChanged = this.onTagsChanged.bind(this);
     }
 
+    public componentDidUpdate(prevProp: IEditorFooterProps) {
```
wbreza: I would like to see a new unit tests for this change.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -325,6 +327,30 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         });
     }
 
+    private async updateProjectTagsFromAsset(asset: IAssetMetadata) {
+        const assetTags = new Set();
+        asset.regions.forEach((region) => region.tags.forEach((tag) => assetTags.add(tag.name)));
+
+        const newTags: ITag[] = this.props.project.tags ? this.props.project.tags.slice() : [];
```
wbreza: can be rewritten as `const newTags = [...this.props.project.tags];`

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -325,6 +327,30 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+        let updateTags = false;
+
+        assetTags.forEach((tag) => {
+            if (!this.props.project.tags || this.props.project.tags.length === 0 ||
+                this.props.project.tags.filter((projectTag) => tag === projectTag.name).length === 0 ) {
```
wbreza: I would use `tags.find(...)` here.  It stops at the first found element vs. traversing the whole array.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -325,6 +327,30 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+                newTags.push({
+                    name: tag,
+                    color: "",
```
wbreza: We should reuse the `tagsInput` method to generate a color for this newly created tag.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -325,6 +327,30 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         });
     }
 
+    private async updateProjectTagsFromAsset(asset: IAssetMetadata) {
```
wbreza: I would like to see a new unit tests validating this behavior in the editor page.

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: You still have to check if null before otherwise you got a spread error (basically null is non iterable).  So in this case slice or spread are really the same but I'll go with spread as it is well spread in the codebase ;-)

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: I just wonder why we have color in project.tags (ITag).  It seams a duplication of info as we don't really have random color or any way to let the user specify a color.  It should just be the correspondence position in the TagColors collection.

wbreza: The user can change the tag color by `Ctrl+Click` on the tag in either the editor or properties page.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -324,6 +328,30 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+                newTags.push({
+                    name: tag,
+                    color: TagColors[newTags.length],
+                });
+                updateTags = true;
+            }
+        });
+
+        if (updateTags) {
+            const newProject = {...this.props.project, tags: newTags};
```
mydiemho: why are we creating a new project? What will happen to the old one?

src/services/assetService.ts
```diff
@@ -200,15 +200,9 @@ export class AssetService {
         const objectArray = await this.getTFRecordObjectArrays(asset);
```
mydiemho: I would rename this `TFRecordMetadata`, objectArray is too generic

src/services/assetService.ts
```diff
@@ -200,15 +200,9 @@ export class AssetService {
...
         const regions: IRegion[] = [];
         const tags: string[] = [];
-        let tagPos = 0;
 
         // Add Regions from TFRecord in Regions
         for (let index = 0; index < objectArray.textArray.length; index++) {
-            tagPos = tags.findIndex((tag) => tag === objectArray.textArray[index]);
```
mydiemho: why are we removing this?

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: If you see just next line you see we are calling the redux action to save the project.  Because of redux immutability it has to be a new project.  The redux/react flow will update back everything.

src/services/assetService.ts
JacopoMangiavacchi: If you look at the code below you see this is not needed anymore.

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: Integrated the test about forcing the tag from the asset metadata directly into the existing unit tests with minimal impact on mocks.  Otherwise I had to duplicate lot of mock and beforeall code with very few changes.

src/react/components/pages/editorPage/editorFooter.test.tsx
```diff
@@ -126,4 +126,24 @@ describe("Footer Component", () => {
...
+
+        wrapper.setProps({tags: [...originalTags, {color: "#808000", name: "NEWTAG"}]});
+
+        await MockFactory.flushUi();
+        wrapper.update();
+
+        const tagChild = wrapper.find("div.tag");
+
+        expect(tagChild.length).toEqual(originalTags.length + 1);
+        expect(tagChild.at(tagChild.length - 1).childAt(0).childAt(1).props().children[1]).toEqual("NEWTAG");
```
wbreza: Is this looking for the new tag in the list?  would `expect(tagChild.last().text()).toEqual("NEWTAG")` work?  This current selector is a bit difficult to follow.

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -196,6 +196,74 @@ describe("Editor Page Component", () => {
         expect(saveProjectSpy).toBeCalledWith(expect.objectContaining(partialProject));
     });
 
+    describe("Editor Page Component Forcing Tag Scenario", () => {
```
wbreza: You likely don't need all of this setup code - It is duplicate from the other tests in the file.

wbreza: I've pushed a commit to simplify. Let me know if you have any questions.

src/common/mockFactory.ts
```diff
@@ -133,7 +133,7 @@ export default class MockFactory {
     /**
      * Creates a mock region
      */
-    public static createMockRegion(id?: string): IRegion {
+    public static createMockRegion(id?: string, forcedNewTag?: string): IRegion {
```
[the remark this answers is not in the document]
JacopoMangiavacchi: Yes, I forgot to remove that changes.  I had this in the previous integrated test when I check for the forcing tag scenario in the existing unit test.  I'm removing this.

src/react/components/pages/editorPage/editorPage.test.tsx
JacopoMangiavacchi: Yes, thank you.  See comments above

src/react/components/pages/editorPage/editorFooter.test.tsx
JacopoMangiavacchi: Yes, it works with ...toMatch()

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -293,12 +293,14 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
-        canvas.style.height = `${contentSource.offsetHeight}px`;
-        this.editor.resize(contentSource.offsetWidth, contentSource.offsetHeight);
+        if (contentSource) {
+            const canvas = this.canvasZone.current;
+            canvas.style.top = `${contentSource.offsetTop}px`;
+            canvas.style.left = `${contentSource.offsetLeft}px`;
+            canvas.style.width = `${contentSource.offsetWidth}px`;
+            canvas.style.height = `${contentSource.offsetHeight}px`;
+            this.editor.resize(contentSource.offsetWidth, contentSource.offsetHeight);
+        }
```
JacopoMangiavacchi: @wbreza this fix the Windows CI build issue but anyway this canvas variable it's not needed at all.  We should remove this.

PR #589, opened 2019-02-15
Removed toolbar items that are not implemented

wbreza:
- Removed zoom in
- Removed zoom out
- Removed pan
- Commented out Copy Rectangle


src/registerToolbar.ts
```diff
@@ -40,13 +29,14 @@ export default function registerToolbar() {
...
     });
 
-    ToolbarItemFactory.register(CopyRectangle, {
-        name: "copyRectangle",
-        tooltip: "Copy Rectangle",
-        icon: "fa-clone",
-        group: "canvas",
-        type: ToolbarItemType.State,
-    });
+    // This is buggy and doesn't fully work yet
```
mydiemho: we should remove it then :D, we should not have commented out code

wbreza: Removed.

PR #593, opened 2019-02-20
props are null when using keyboard shortcuts

mydiemho:
1. `this` is null

### DIFF
--- a/src/react/components/common/assetPreview/videoAsset.tsx
+++ b/src/react/components/common/assetPreview/videoAsset.tsx
@@ -66,6 +66,15 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
         duration: 0,
     };
 
+    constructor(props, context) {
+        super(props, context);
+
+        this.moveNextTaggedFrame = this.moveNextTaggedFrame.bind(this);
+        this.movePreviousTaggedFrame = this.movePreviousTaggedFrame.bind(this);
+        this.moveNextExpectedFrame = this.moveNextExpectedFrame.bind(this);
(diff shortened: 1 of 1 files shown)


src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -66,6 +66,15 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
         duration: 0,
     };
 
+    constructor(props, context) {
+        super(props, context);
+
+        this.moveNextTaggedFrame = this.moveNextTaggedFrame.bind(this);
```
mydiemho: this fix make sure `this` is not null.  however, `additionalVideoSetting` is still null

tbarlow12: Questions for clarification: 
1. Where is `this` null? In each of these functions?
2. Is `additionalSettings` the null value? Or is it `additionalSettings.videoSettings`? It looks like those are coming from the props, which means their value would be set upstream

tbarlow12: Also, do you have any unit tests to validate the new behavior?

wbreza:
@mydiemho - The main issue here is related to how the child `AssetPreview` is rendered within the `Canvas` component.

src/react/components/common/assetPreview/videoAsset.tsx
mydiemho: 1. `this` is null in each of the functions when using keyboard, they're fine when using the on screen button
2. `additionalSettings` is the null value, you can try it out by adding breakpoint.

mydiemho:
@wbreza is going to pick this up

PR #598, opened 2019-02-22
Fix failing tests in export video branch

wbreza:
Fixes failing tests in Phi's export video branch


src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -243,7 +243,6 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
             const rootAsset = this.props.asset.parent || this.props.asset;
             const childPath = `${rootAsset.path}#t=${state.currentTime}`;
             const childAsset = AssetService.createAssetFromFilePath(childPath);
-            childAsset.name = `${rootAsset.name + "." + rootAsset.format}#t=${state.currentTime}`;
```
wbreza: I believe this line was causing a majority of the test failures.  I don't see why we need to override the name of the video frame.  When exporting it should use the generated ID.

PR #599, opened 2019-02-22
(WIP)feature:Add support for async forEach/map

wbreza:
Adds support for processing/batching array operations in parallel.

### DIFF
--- a/src/common/array.ts
+++ b/src/common/array.ts
@@ -1,0 +1,56 @@
+import Guard from "./guard";
+
+/**
+ * Processes items in the array within the specified batch size (default: 5)
+ * @param this The array to process
+ * @param action The action to perform on each item in the array
+ * @param batchSize The batch size for actions to perform in parallel (default: 5)
+ */
+export async function forEachAsync<T>(
+    this: T[],
+    action: (item: T) => Promise<void>,
+    batchSize: number = 5): Promise<void> {
+    Guard.null(this);
+    Guard.null(action);
+
+    const all: T[] = [...this];
+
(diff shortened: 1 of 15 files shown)


src/common/array.ts
```diff
@@ -0,0 +1,59 @@
+import Guard from "./guard";
+
+/**
+ * Processes items in the array within the specified batch size (default: 5)
+ * @param this The array to process
+ * @param action The action to perform on each item in the array
+ * @param batchSize The batch size for actions to perform in parallel (default: 5)
+ */
+export async function arrayForEachAsync<T>(
+    this: T[],
```
tbarlow12: Is this similar to extensions in C#?

src/common/htmlFileReader.ts
```diff
@@ -86,21 +86,24 @@ export default class HtmlFileReader {
...
 
     public static async getAssetFrameImage(asset: IAsset) {
         return new Promise((resolve, reject) => {
-            let video: HTMLVideoElement;
-            let refresh = true;
-            if (asset.parent.name in this.videoAssetFiles) {
-                video = this.videoAssetFiles[asset.parent.name];
-                refresh = false;
-            } else {
+            const cachingEnabled = false;
```
tbarlow12: How does `cachingEnabled` ever get set to true?

PIC123: So right now it doesn't get reset, it's just for development as a toggle while we see how we can add in the caching, but for now the exporting works fine without the caching since the batching can help optimize things

PR #600, opened 2019-02-22
feat: Locked tags for tagging regions [AB#16986]

tbarlow12:


wbreza:
@tbarlow12 - PR needs scope + description

src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -219,31 +222,166 @@ describe("Editor Canvas", () => {
...
+        canvas.applyTag(newTag);
+
+        const original = getAssetMetadata();
+        const expected = cloneWithUpdatedRegionTags(original, "test1", [newTag]);
+
+        expect(onAssetMetadataChanged).toBeCalledWith(expected);
+        expect(wrapper.state().currentAsset.regions[0].tags).toEqual([newTag]);
+    });
+
+    it("Removes old locked tag from selected region", () => {
```
wbreza: Is description accurate - it doesn't appear that this test is removing any tags, but applying new ones.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -119,10 +140,11 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
             this.props.selectedAsset.asset.size.width,
             this.props.selectedAsset.asset.size.height,
         );
+        const lockedTags = this.props.lockedTags;
         const newRegion = {
             id,
             type: this.editorModeToType(this.props.editorMode),
-            tags: [],
+            tags: (lockedTags) ? lockedTags : [],
```
wbreza: you can simplify this as `lockedTags || []` If `lockedTags` is defined it will use its value, otherwise an empty array.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -194,18 +219,34 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
     /**
      * Method called when deleting a region from the editor
      * @param {string} id the id of the deleted region
-     * @param {boolean} multiselection boolean whether multiselect is active
+     * @param {boolean} multiselect boolean whether region was selected with multiselection
      * @returns {void}
      */
     private onRegionSelected = (id: string, multiselect: boolean) => {
-        const region = this.state.currentAsset.regions.find((region) => region.id === id);
+        if (!id) {
```
wbreza: How can `onRegionSelected` get called without an id?

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -311,17 +352,11 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
-     * if already contained in tags. Update tags in CanvasTools editor
-     * @param region Region to add or remove tag
-     * @param tag Tag to add or remove from region
-     */
-    private toggleTagOnRegion = (region: IRegion, tag: string) => {
-        CanvasHelpers.toggleTag(region.tags, tag);
-        this.editor.RM.updateTagsById(region.id, CanvasHelpers.getTagsDescriptor(this.props.project.tags, region));
-        const updatedRegions = this.state.currentAsset.regions.map((r) => (r.id === region.id) ? region : r);
-        const updatedSelectedRegions = this.state.currentAsset.regions.map((r) => (r.id === region.id) ? region : r);
+    private updateRegions = (updates: IRegion[], updatedSelectedRegions: IRegion[]) => {
```
wbreza: What is the different between updates vs updatedSelectedRegions?  Let's add some more docs to make this easier to understand.

src/react/components/pages/editorPage/canvasHelpers.ts
```diff
@@ -27,6 +27,89 @@ export default class CanvasHelpers {
...
+    public static toggleRegion(regions: IRegion[], region: IRegion): void {
+        const index = regions.findIndex((r) => r.id === region.id);
+        if (index === -1) {
+            regions.push(region);
+        } else {
+            regions.splice(index, 1);
+        }
+    }
+
+    /**
```
wbreza: Not sure these add much value (finde, findRegion, findIndex) since the replaced code is essentially the same... Thoughts?

src/react/components/pages/editorPage/canvasHelpers.ts
```diff
@@ -27,6 +27,89 @@ export default class CanvasHelpers {
...
+    public static addIfMissing(tags: string[], tag: string): void {
+        if (!CanvasHelpers.find(tags, tag)) {
+            tags.push(tag);
+        }
+    }
+
+    /**
+     * Adds all target tags if missing from tags
+     */
+    public static addAllIfMissing(tags: string[], targets: string[]): void {
```
wbreza: What is `targets` here?  Region ids? tags?

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -184,8 +191,20 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
      * Called when a tag from footer is clicked
      * @param tag Tag clicked
      */
-    public onTagClicked = (tag: ITag): void => {
-        this.canvas.current.applyTag(tag.name);
+    private onTagClicked = (tag: ITag): void => {
```
wbreza: I like that `lockedTags` is passed into the canvas via props - Is there a reason we're not doing the same for `selectedTag`?  Let's reduce usage of the canvas ref to as little as possible.

src/react/components/pages/editorPage/canvasHelpers.ts
tbarlow12: True. I was using it a lot more when I was refactoring, and thought I was saving some time. But I'll take these pieces out

src/react/components/pages/editorPage/canvasHelpers.ts
tbarlow12: Tags. I'll change the name and add docs

src/react/components/pages/editorPage/editorPage.tsx
tbarlow12: I struggled with this for a long time. The main piece that was a blocker was how to differentiate changes. For example, if I select `Tag1` to add it to a tag, I can make the change in `componentDidUpdate` no problem. However, if I change my mind and want to remove `Tag1` from the region, I click it again, setting the state to exactly what it was before and is not noticed as a change in the `componentDidUpdate` function. Happy to chat about other possible solutions though

src/react/components/pages/editorPage/canvas.tsx
tbarlow12: This is a CT bug I've mentioned to Konstantin. When a region is deselected, id is undefined. Not sure why, this is a blocker for multiselect to work. Hoping he can get a fix in for this

src/react/components/pages/editorPage/canvas.tsx
tbarlow12: Good catch

src/react/components/pages/editorPage/canvas.test.tsx
tbarlow12: You're right, the logic is a bit off. I meant to apply one of the locked tags and have it missing from the lockedTags array, simulating the action of having regions selected and deslecting a locked tag. I'll make the fix

src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -219,31 +222,166 @@ describe("Editor Canvas", () => {
...
+        const onAssetMetadataChanged = jest.fn();
+        wrapper.setProps({ onAssetMetadataChanged });
+        const canvas = wrapper.instance();
+
+        canvas.editor.onRegionSelected("test1", null);
+
+        const newTag = "newTag";
+        canvas.applyTag(newTag);
+
+        const original = getAssetMetadata();
```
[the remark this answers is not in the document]
tbarlow12: I actually just want fake data here. Original is just the name I picked for "asset before I start simulating changes on it"

src/react/components/pages/editorPage/editorPage.tsx
wbreza: We can skip for now and address this later.

PR #601, opened 2019-02-22
fix:Resolves issue where user is unable to create new project

wbreza:
Fixes a bug where when attempting to create a new project after already have opened a project it does not close the existing project before navigating the user to the new project screen. 

Also resolves issue where security token picker component was not rendering in in project settings form.

Resolves AB#17172, AB#17177, AB#17179, AB#17180


wbreza:
@JacopoMangiavacchi - It is by design that we don't re-load previous project if you cancel on new project screen.

PR #604, opened 2019-02-23
fix: Correct bounding box values on region move

wbreza:
Corrects an issue where bounding box position was not being updated correctly when a region was moved within the canvas editor.  Also randomized region test data to ensure we are tests are more accurate

### DIFF
--- a/src/common/mockFactory.ts
+++ b/src/common/mockFactory.ts
@@ -27,6 +27,7 @@ import { AssetService } from "../services/assetService";
 import { SelectionMode } from "vott-ct/lib/js/CanvasTools/Selection/AreaSelector";
 import { Point2D } from "vott-ct/lib/js/CanvasTools/Core/Point2D";
 import { RegionDataType, RegionData } from "vott-ct/lib/js/CanvasTools/Core/RegionData";
+import { randomIntInRange } from "./utils";
 
 export default class MockFactory {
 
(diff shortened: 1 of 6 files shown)


src/common/mockFactory.ts
```diff
@@ -577,32 +545,73 @@ export default class MockFactory {
...
+        };
+
+        return {
+            id,
             boundingBox: {
-                height: 100,
-                width: 100,
-                left: 0,
-                top: 0,
+                left: origin.x,
```
tbarlow12: Nice

src/common/mockFactory.ts
```diff
@@ -901,4 +910,8 @@ export default class MockFactory {
                 return StorageType.Other;
         }
     }
+
+    private static randomNumber(min: number, max: number) {
```
tbarlow12: Not that this matters much since it's so simple, but this already exists in the `utils` file

wbreza: Thanks - will convert to use that method.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -168,6 +168,12 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
 
         if (movedRegion) {
             movedRegion.points = scaledRegionData.points;
+            movedRegion.boundingBox = {
+                height: scaledRegionData.height,
```
wbreza: This was the core bug fix.

PR #606, opened 2019-02-25
docs: update contribution guidelines for changelog

pjlittle:
Contribution guidelines now reflect new git commit message conventions, to support change log generation.

AB#17149

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,28 @@
+# Contributing to VoTT
 We welcome [issues](https://github.com/Microsoft/VoTT/issues) and [pull requests](https://github.com/Microsoft/VoTT/pulls) into the project. We ask that you follow these simple guidelines:
 
-### Issues
+<!-- toc -->
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -15,11 +32,67 @@ We welcome [issues](https://github.com/Microsoft/VoTT/issues) and [pull requests
...
+<BLANK LINE>
+<footer>
+```
+
+### Header
+#### Type
+Must be one of the following:
+
+* **build**: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
+* **ci**: Changes to our CI configuration files and scripts (example scopes: Circle, BrowserStack, SauceLabs)
```
luisamiranda: should we make these example scopes specific to what we're using? we are not using gulp, broccoli, CircleCI, etc.

pjlittle: ah, good point... we elected to not include scope.  I'll take this out.

PR #609, opened 2019-02-25
feat: Add Project version info

JacopoMangiavacchi:
- Add Project version to project file
- Add Project version to asset metadata file
- Show Project version in App Settings Page

Implement [#AB17100]

### DIFF
--- a/src/models/applicationState.ts
+++ b/src/models/applicationState.ts
@@ -219,6 +219,7 @@ export interface IAsset {
 export interface IAssetMetadata {
     asset: IAsset;
     regions: IRegion[];
+    version: string;
 }
 
 /**
--- a/src/models/applicationState.ts
+++ b/src/models/applicationState.ts
@@ -98,6 +98,7 @@ export interface IAppSettings {
 export interface IProject {
     id: string;
     name: string;
+    version: string;
     securityToken: string;
     description?: string;
     tags: ITag[];
--- a/src/react/components/pages/appSettings/appSettingsPage.tsx
+++ b/src/react/components/pages/appSettings/appSettingsPage.tsx
@@ -62,7 +62,7 @@ export default class AppSettingsPage extends React.Component<IAppSettingsProps> 
                     onCancel={this.onFormCancel} />
                 <div className="app-settings-page-sidebar p-3 bg-lighter-1">
                     <div className="my-3">
-                        <p>{strings.appSettings.version.description + packageJson.version}</p>
+                        <p>{`${strings.appSettings.version.description} ${packageJson.version}`}</p>
                     </div>
                     <div className="my-3">
                         <p>{strings.appSettings.devTools.description}</p>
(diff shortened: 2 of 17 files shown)


src/models/applicationState.ts
```diff
@@ -98,6 +98,7 @@ export interface IAppSettings {
 export interface IProject {
```
wbreza: What do you think about defining the `version` as optional - So it makes defining the objects easier from the consumer side and then the services/reducers job is to set it when it changes.

src/services/assetService.test.ts
```diff
@@ -150,6 +151,7 @@ describe("Asset Service", () => {
                     state: AssetState.Tagged,
                 },
                 regions: [],
+                version: "",
             };
```
wbreza: I would expect this test to fail right now - If we are setting version to empty string the `assetService.save(...)` should be applying the version from the package.json therefore making the equality comparison fail below.

src/redux/actions/projectActions.ts
```diff
@@ -70,7 +71,9 @@ export function saveProject(project: IProject)
             throw new AppError(ErrorCode.SecurityTokenNotFound, "Security Token Not Found");
         }
 
-        const savedProject = await projectService.save(project, projectToken);
+        const newProject = {...project, version: packageJson.version};
```
wbreza: Let's make sure we have a new test the validates the version is set when `saveProject` is called.

src/models/applicationState.ts
JacopoMangiavacchi: It's "optional" to read of course but I think it should be mandatory to always persist project (and asset) with this new value.  If we mark this optional we're not guaranteed.

src/services/assetService.test.ts
JacopoMangiavacchi: Not really. The assetService.save() method just save an asset metadata it receives in input.  In that specific test the goal was just to check this method that don't really care about the value of asset metadata. The thing that that property is not optional just enforce that at least it as some value and an empty string is still a value.   Btw real project version value is enforced in the redux.  All the test that use the redux save project has been updated in this PR asserting for the valid project version.

src/redux/actions/projectActions.ts
JacopoMangiavacchi: Added test for both saveProject and saveAssetMetadata

src/react/components/pages/appSettings/appSettingsPage.tsx
```diff
@@ -60,6 +61,9 @@ export default class AppSettingsPage extends React.Component<IAppSettingsProps>
                     onSubmit={this.onFormSubmit}
                     onCancel={this.onFormCancel} />
                 <div className="app-settings-page-sidebar p-3 bg-lighter-1">
+                    <div className="my-3">
+                        <p>{strings.appSettings.version.description + packageJson.version}</p>
```
wbreza: Consider using string interpolation vs appending string.

PR #610, opened 2019-02-25
feat: Add ability to import v1 projects

elizabethhalper:
This adds the ability for a user to import their v1 projects into the v2 app and convert the project to v2 format. Once the user imports the project, their changes to the project will be saved in the new v2 .vott file.


src/models/v1Models.ts
```diff
@@ -0,0 +1,55 @@
+import { ExportAssetState } from "../providers/export/exportProvider";
+import { IPoint } from "./applicationState";
+/**
+ * @name - V1 Project
+ * @description - Defines the structure of a version 1 Project
+ * @member frames - Dictionary of all frame objects in the project (filename: list of regions in that file)
+ * @member framerate - Rate at which a video is stepped through
+ * @member inputTags - Comma delimited list of all tags in the project
+ * @member suggestiontype - TODO: Add description
```
pjlittle: description for these two

src/common/mockFactory.ts
```diff
@@ -207,6 +207,54 @@ export default class MockFactory {
...
 
+    /**
+     * Creates fake IV1Project
+     * @param name Name of project.
+     */
+    public static createTestV1Project(): IV1Project {
+        const regions = MockFactory.createTestV1Regions(10);
+
+        return {
+            frames: {"testFrame.jpg": regions},
```
JacopoMangiavacchi: Are we testing importing V1 projects with more than one assets/frames ?

src/common/mockFactory.ts
```diff
@@ -207,6 +207,54 @@ export default class MockFactory {
...
+                y2: 11 },
+            points: [],
+            UID: count.toString(),
+            id: 2,
+            type: "rect",
+            tags: [],
+            name: 1,
+        };
+
+        regions.push(testRegion);
```
JacopoMangiavacchi: Also here, maybe we want to test importing assets with more than a region

src/services/assetService.ts
```diff
@@ -179,7 +177,7 @@ export class AssetService {
     public async getAssetMetadata(asset: IAsset): Promise<IAssetMetadata> {
         Guard.null(asset);
 
-        const fileName = `${asset.id}${constants.assetMetadataFileExtension}`;
+        const fileName = `${asset.id}-asset.json`;
```
JacopoMangiavacchi: Why we're not using anymore constants.assetMetadataFileExtension ?

src/services/importService.ts
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * @name - Import Service
+ * @description - Functions for importing v1 projects to v2 application
+ */
+export default class ImportService implements IImportService {
+    private actions: IProjectActions;
+    constructor(actions: IProjectActions) {
+        this.actions = actions;
+    }
+    public async convertProject(projectInfo: IFileInfo): Promise<IProject> {
```
JacopoMangiavacchi: You may want to add docs for this public method

src/common/mockFactory.ts
elizabethhalper: Got it! Added multiple assets and multiple regions to the test

src/common/localization/en-us.ts
```diff
@@ -29,6 +29,12 @@ export const english: IAppStrings = {
             title: "Delete Project",
             confirmation: "Are you sure you want to delete project",
         },
+        importProject: {
+            title: "Import Project",
+            confirmation: "Are you sure you want to convert project",
```
tbarlow12: What was the purpose of separating these two out again? I feel like we talked about it, but I can't remember

src/common/mockFactory.ts
```diff
@@ -268,6 +268,59 @@ export default class MockFactory {
...
+
+    /**
+     * Creates fake IV1Region
+     */
+    public static createTestV1Regions(count: number = 3): IV1Region[] {
+        const regions: IV1Region[] = [];
+
+        for (let i = 0; i < count; i++) {
+            const testRegion = {
+                x1: 1,
```
tbarlow12: nit: Any way we could randomize these dimensions? Just to avoid magic numbers if possible and keep the tests honest

tbarlow12: There are other examples within the `MockFactory` of this kind of stuff

src/react/components/pages/homepage/homePage.test.tsx
```diff
@@ -4,26 +4,37 @@ import { Provider } from "react-redux";
...
 import CondensedList from "../../common/condensedList/condensedList";
+import Confirm, { IConfirmProps } from "../../common/confirm/confirm";
 import FilePicker, { IFilePickerProps } from "../../common/filePicker/filePicker";
 import HomePage, { IHomepageProps, IHomepageState } from "./homePage";
 
 jest.mock("../../common/cloudFilePicker/cloudFilePicker");
 import { CloudFilePicker, ICloudFilePickerProps } from "../../common/cloudFilePicker/cloudFilePicker";
 
 jest.mock("../../../../services/projectService");
+jest.mock("../../../../services/importService");
```
tbarlow12: Does this work? I was under the impression that the import statement needed to come after the mock statement. But that'd be great if I'm wrong!

src/react/components/pages/homepage/homePage.test.tsx
```diff
@@ -4,26 +4,37 @@ import { Provider } from "react-redux";
...
 
 describe("Homepage Component", () => {
     let store: Store<IApplicationState> = null;
     let props: IHomepageProps = null;
     let wrapper: ReactWrapper = null;
     let deleteProjectSpy: jest.SpyInstance = null;
     let closeProjectSpy: jest.SpyInstance = null;
+    // let importServiceMock: jest.Mocked<typeof ImportService> = null;
```
tbarlow12: Remove commented out line

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -44,10 +56,29 @@ export default class HomePage extends React.Component<IHomepageProps> {
     public state: IHomepageState = {
         cloudPickerOpen: false,
     };
-
     private filePicker: React.RefObject<FilePicker> = React.createRef();
     private deleteConfirm: React.RefObject<Confirm> = React.createRef();
     private cloudFilePicker: React.RefObject<CloudFilePicker> = React.createRef();
+    private importConfirm: React.RefObject<Confirm> = React.createRef();
+
+    constructor(props: IHomepageProps, context) {
```
tbarlow12: You should be able to do this without the constructor. You can define state inline like:
```
export default class HomePage extends React.Component<IHomepageProps> {

    state = {
        cloudPickerOpen: false
    }
```
Also, on lines 59-62, you're already creating the Ref instance, so you don't need the creation inside the constructor. Nit: you can also just add the type in that first instantiation

And for the binding of functions, since you're using `function x = () => {` syntax, it automatically binds for you :)

src/common/localization/en-us.ts
elizabethhalper: This is just so it reads smoothly. I have to put the name of the project in the middle of the alert but didn't want to hard-code any of the dialogue in english

src/common/mockFactory.ts
elizabethhalper: Yeah I can do that. I was originally testing the entire region, but I can just test that it exists with a specific UID than testing the whole object.

src/react/components/pages/homepage/homePage.test.tsx
elizabethhalper: It does work, but it's not as clean/pretty. I changed it so it goes with convention :)

src/common/localization/en-us.ts
wbreza: @elizabethhalper  - We can use string interpolation inside the localized strings.  See `projectSettings.messages.saveSuccess` for an example.

src/react/components/common/confirm/confirm.tsx
```diff
@@ -16,6 +16,7 @@ export interface IConfirmProps extends IMessageBoxProps {
     cancelButtonText?: string;
     confirmButtonColor?: string;
     cancelButtonColor?: string;
+    className?: string;
```
wbreza: className should be intrinsic on all react components... You may just need to ensure the component interface is extending the `React.Props` interface.

src/react/components/pages/homepage/homePage.tsx
elizabethhalper: Got it! I had added the constructor to circumvent an issue that was solved by another change so I switched it back to how it was before.

src/react/components/common/confirm/confirm.tsx
elizabethhalper: @wbreza what if it already extends the `React.Component` interface? Can't it only extend one of these?

src/common/localization/en-us.ts
elizabethhalper: done!

PR #611, opened 2019-02-25
feat: Enable copy rectangle functionality in editor

PIC123:
This adds the copy rectangle functionality of canvas tools to the VOTT editor. When a region is selected on the canvas, the user can select the copyRect button to switch modes and create regions, using the shape of the previously selected region as a template.


src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -75,12 +74,21 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
         if (this.props.selectionMode !== prevProps.selectionMode) {
-            this.editor.AS.setSelectionMode(this.props.selectionMode, null);
+            let  options = null;
+            if (this.props.selectionMode === SelectionMode.COPYRECT) {
+                const selectedRegion = this.getSelectedRegions(); // this.state.selectedRegions[0];
+                if (selectedRegion) {
+                    options = {template: new Rect(selectedRegion[0].boundingBox.width,
+                                                selectedRegion[0].boundingBox.height)};
+                } else {
+                    throw new Error("No region selected to copy");
```
kichinsky: If no template is provided, CanvasTools will use a default one (20x20 square). So such use case is not considered as an error.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -75,12 +74,21 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
         }
 
         if (this.props.selectionMode !== prevProps.selectionMode) {
-            this.editor.AS.setSelectionMode(this.props.selectionMode, null);
+            let  options = null;
+            if (this.props.selectionMode === SelectionMode.COPYRECT) {
+                const selectedRegion = this.getSelectedRegions(); // this.state.selectedRegions[0];
+                if (selectedRegion) {
+                    options = {template: new Rect(selectedRegion[0].boundingBox.width,
+                                                selectedRegion[0].boundingBox.height)};
```
kichinsky: It should be just `template`:
{ 
  mode: SelectionMode.COPYRECT,
  template: new Rect(...),
}

src/react/components/pages/editorPage/canvas.tsx
kichinsky: Just in case you would like to cover it with a test: set the AreaSelector.DefaultTemplateSize first to make it controllable. It defaults to `new Rect(20, 20)`

PR #612, opened 2019-02-25
fix: Fixes issue where regions cannot be deleted

wbreza:
Fixes the issue where the last region in an asset is unable to be deleted.  When all the regions on an asset are deleted the asset transitions back into the "Visited" state.  Therefore, when the asset metadata is saved is is not persisted back to the file system since it is no longer in the "Tagged" state.  The fix is to instead of saving the asset metadata file it should be deleted from the storage provider since the asset metadata file doesn't not provide any additional metadata outside of region data.

Resolves AB#17174

### DIFF
--- a/src/electron/providers/storage/localFileSystem.ts
+++ b/src/electron/providers/storage/localFileSystem.ts
@@ -78,15 +78,17 @@ export default class LocalFileSystem implements IStorageProvider {
     public deleteFile(filePath: string): Promise<void> {
         return new Promise<void>((resolve, reject) => {
             const exists = fs.existsSync(path.normalize(filePath));
-            if (exists) {
-                fs.unlink(filePath, (err) => {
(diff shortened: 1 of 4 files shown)


src/electron/providers/storage/localFileSystem.ts
```diff
@@ -78,15 +78,17 @@ export default class LocalFileSystem implements IStorageProvider {
...
             const exists = fs.existsSync(path.normalize(filePath));
-            if (exists) {
-                fs.unlink(filePath, (err) => {
-                    if (err) {
-                        return reject(err);
-                    }
-
-                    resolve();
-                });
+            if (!exists) {
```
wbreza: There was a bug in the deleteFile where the promise would never resolve/reject if the filePath didn't exist.

PR #617, opened 2019-02-26
feat: add hotkeys for all editor toolbar items and simplify tag input hotkeys (remove ctrl+)

luisamiranda:
This feature aims to make it easier for a user to keep one hand on the keyboard and one on the mouse for more efficient tagging of large quantities of images. Previous Asset (W) and Next Asset (S) already exists. This feature adds hotkeys to the editor toolbar as follows:
- Select (V) (following common hotkey convention)
- Draw Rectangle (R)
- Draw Polygon (P)
- Save Project (Ctrl + S)
- Export Project (Ctrl + E)
In addition, tag hotkeys have been simplified to numbers 1 through 0, instead of Ctrl + 1, etc.

AB#17096

### DIFF
--- a/src/react/components/common/keyboardManager/keyboardRegistrationManager.ts
+++ b/src/react/components/common/keyboardManager/keyboardRegistrationManager.ts
@@ -1,6 +1,6 @@
 import Guard from "../../../../common/guard";
 import { KeyEventType } from "./keyboardManager";
-
+// FIXME: should KeyEventType enum in keyboardManager be moved here to avoid circular dependency?
 /**
  * A map of keyboard event registrations
  */
--- a/src/react/components/common/keyboardManager/keyboardRegistrationManager.ts
+++ b/src/react/components/common/keyboardManager/keyboardRegistrationManager.ts
@@ -1,13 +1,5 @@
 import Guard from "../../../../common/guard";
-
-/**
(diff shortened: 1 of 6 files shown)


src/react/components/common/keyboardManager/keyboardRegistrationManager.ts
```diff
@@ -1,6 +1,6 @@
 import Guard from "../../../../common/guard";
 import { KeyEventType } from "./keyboardManager";
-
+// FIXME: should KeyEventType enum in keyboardManager be moved here to avoid circular dependency?
```
wbreza: Are you getting warnings / errors about a circular dependency?

luisamiranda: no, it's not technically circular, but these two files import from each other unnecessarily. The should import in one direction. I can remove the comment or just move the enum from one file to another.

luisamiranda: I can also just write it up as a separate "style" bug.

wbreza: You're going to have the same case either way... Both files use the enum, but it is only ever defined in 1 of them.  Let's remove the comment and keep as is.

PR #619, opened 2019-02-27
fix: User can navigate between key frames using keyboard accelerators

wbreza:
Resolves the issue where keyboard accelerators to navigate between key frames was throwing an error.  Main issue was that multiple keyboard bindings were being registered for the hot keys causing weird behavior issues.  Keyboard bindings are now only registered for the main video player control inside the canvas.

Resolves AB#17130

### DIFF
--- a/src/react/components/common/assetPreview/videoAsset.tsx
+++ b/src/react/components/common/assetPreview/videoAsset.tsx
@@ -73,40 +73,38 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
                 autoPlay={autoPlay}
                 src={videoPath}>
                 <BigPlayButton position="center" />
-                <ControlBar autoHide={false}>
-                    <CustomVideoPlayerButton order={1.1}
-                        keyboardEnabled={!!this.props.autoPlay}
-                        accelerators={["ArrowLeft", "a", "A"]}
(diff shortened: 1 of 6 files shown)


src/react/components/common/assetPreview/videoAsset.tsx
```diff
@@ -73,36 +73,38 @@ export class VideoAsset extends React.Component<IVideoAssetProps> {
...
-                        <i className="fas fa-step-backward"></i>
-                    </CustomVideoPlayerButton>
-                    <CustomVideoPlayerButton order={8.2}
-                        accelerators={["e", "E"]}
-                        tooltip={strings.editorPage.videoPlayer.nextTaggedFrame.tooltip}
-                        onClick={this.moveNextTaggedFrame}>
-                        <i className="fas fa-step-forward"></i>
-                    </CustomVideoPlayerButton>
-                </ControlBar>
+                {autoPlay &&
```
mydiemho: will autoPlay ever be false? it's not something user can set right?  When I navigate away from a video and come back, it always autoplay

wbreza: Right now autoPlay is `false` on the sidebar videos and `true` for the main canvas video.

PR #620, opened 2019-02-27
fix: Delete asset metadata files when project is deleted

wbreza:
Resolves the issue where orphaned asset metadata files still exist on the target connection.

Resolves AB#17101

### DIFF
--- a/src/services/projectService.ts
+++ b/src/services/projectService.ts
@@ -1,6 +1,7 @@
+import _ from "lodash";
 import shortid from "shortid";
 import { StorageProviderFactory } from "../providers/storage/storageProviderFactory";
-import { IProject, ISecurityToken, AppError, ErrorCode } from "../models/applicationState";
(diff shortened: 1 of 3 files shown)


src/services/projectService.ts
```diff
@@ -44,54 +45,48 @@ export default class ProjectService implements IProjectService {
...
+        const storageProvider = StorageProviderFactory.createFromConnection(project.targetConnection);
+        await this.saveExportSettings(project);
+        project = encryptProject(project, securityToken);
+
+        await storageProvider.writeText(
+            `${project.name}${constants.projectFileExtension}`,
+            JSON.stringify(project, null, 4),
+        );
+
+        return project;
```
elizabethhalper: why get rid of try/catch block?

pjlittle:
:+1:

src/services/projectService.ts
wbreza: The try/catch was only there to facilitate rejecting the promise in an error scenario, otherwise the exception would bubble by default.  No reason to use Promises when we can use async/await.

PR #621, opened 2019-02-27
ci: make sure build fail if codecov token is missing

mydiemho:
also update README to point ADO badge to new CI pipeline

have to create a new pipeline using the new YAML wizard
so that GitHub Pipelines App is used instead of webhook.

This fix the issue where re-running the ADO build from the
`Check` tab in GitHub doesn't kick off actual build in ADO.


pjlittle:
looks good.

PR #624, opened 2019-02-28
feat: Copy, Cut, Paste and Clear [AB#16720]

tbarlow12:
- Added generic `Clipboard` wrapper that calls clipboard async API
- Refactor toolbar registration to use generic `ToolbarItem` unless necessary (save and export still need it) and refactored tests to reflect
- Added public `copyRegions`, `cutRegions`, `pasteRegions` and `clearRegions` definitions to `Canvas`
- Added toolbar buttons to call regions management functions with associated hot keys
    - `Ctrl+c` = copy
    - `Ctrl+x` = cut
    - `Ctrl+v` = paste
    - `Ctrl+e` = clear


src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { mount, ReactWrapper } from "enzyme";
+import { mount, ReactWrapper, shallow } from "enzyme";
```
JacopoMangiavacchi: It doesn't seams to me that shallow is used

src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -12,10 +12,19 @@ import Canvas, { ICanvasProps, ICanvasState } from "./canvas";
...
 
 jest.mock("vott-ct/lib/js/CanvasTools/CanvasTools.Editor");
+import { Editor } from "vott-ct/lib/js/CanvasTools/CanvasTools.Editor";
 
 jest.mock("vott-ct/lib/js/CanvasTools/Region/RegionsManager");
+import { RegionsManager } from "vott-ct/lib/js/CanvasTools/Region/RegionsManager";
 
 describe("Editor Canvas", () => {
+
+    function createTestRegionData() {
```
PIC123: So it looks like you aren't using this method, but are using the one defined in mock factory, so you can probably get rid of this

PR #625, opened 2019-02-28
fix: Resolves typescript tsc compile issues for v3.1.6

wbreza:
When running on typescript < 3.2.x there were several TS errors including:

- Warnings about --isolated modules due to import of package.json
- Warnings about `Promise<{}>` vs `Promise<void>`
- HtmlFileReader usage was inconsistent between use of `ArrayBuffer` vs `UInt8Array`

### DIFF
--- a/src/common/appInfo.ts
+++ b/src/common/appInfo.ts
@@ -1,0 +1,19 @@
+// tslint:disable-next-line:no-var-requires
+const packageJson = require("../../package.json");
+
+/**
+ * Defines the application information
+ */
+export interface IAppInfo {
+    /** The app name */
+    name: string;
+    /** The app version */
+    version: string;
+    /** The app description */
+    description: string;
+}
+
+/**
+ * Gets current application info
+ */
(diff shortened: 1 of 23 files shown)


src/common/appInfo.ts
```diff
@@ -0,0 +1,19 @@
+// tslint:disable-next-line:no-var-requires
+const packageJson = require("../../package.json");
```
wbreza: Instead of explicitly importing `package.json` in multiple places, refactored this into a const that can be easily imported by other modules.

src/common/htmlFileReader.ts
```diff
@@ -94,16 +94,15 @@ export default class HtmlFileReader {
      * Downloads the binary array from the asset path
      * @param asset The asset to download
      */
-    public static async getAssetArray(asset: IAsset): Promise<Uint8Array> {
+    public static async getAssetArray(asset: IAsset): Promise<ArrayBuffer> {
```
JacopoMangiavacchi: Faster, I like it

PR #626, opened 2019-02-28
fix: Move last visited tag settings to reducer to fix delete file assets bug [#AB17101]

JacopoMangiavacchi:
- Move last visited tag settings from editor Page to project Redux action
- Removed non necessary filter on selecting asset files to delete

Fix bug [#AB17101]

### DIFF
--- a/src/services/projectService.ts
+++ b/src/services/projectService.ts
@@ -75,7 +75,6 @@ export default class ProjectService implements IProjectService {
 
         // Delete all asset metadata files created for project
         const deleteFiles = _.values(project.assets)
-            .filter((asset) => asset.state === AssetState.Tagged)
             .map((asset) => storageProvider.deleteFile(`${asset.id}${constants.assetMetadataFileExtension}`));
 
         await Promise.all(deleteFiles);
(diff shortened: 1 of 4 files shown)


src/services/projectService.ts
```diff
@@ -75,7 +75,6 @@ export default class ProjectService implements IProjectService {
 
         // Delete all asset metadata files created for project
         const deleteFiles = _.values(project.assets)
-            .filter((asset) => asset.state === AssetState.Tagged)
             .map((asset) => storageProvider.deleteFile(`${asset.id}${constants.assetMetadataFileExtension}`));
```
elizabethhalper: Why did filtering based on state cause the files not to be deleted? Weren't the files that are saved only corresponding to assets that are tagged?

JacopoMangiavacchi: We're cleaning up all metadata files when deleting the project, that filter is not strictly needed IMHO even if you're right, they should be all in the tagged state but I saw something strange once with an asset with regions but no tags that was there with a different state.

So I thought it's better to clean any metadata file we can when deleting the project.

PR #629, opened 2019-03-01
fix: Fixes issue where user is unable to navigate to new project screen

wbreza:
Resolves the issue where user is unable to navigate to new project screen from the homepage.  This also addresses other strange behavior related to routing in the app.

- User being navigated to the homepage after creating a connection
- Refresh application button not working

Resolves AB#17258, 17263, 17058

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -1,29 +1,32 @@
 const path = require("path");
 
 module.exports = {
-  target: "electron-main",
-  entry: "./src/electron/main.ts",
-  module: {
-    rules: [
-      {
-        test: /\.ts?$/,
-        use: [{
-          loader: "ts-loader",
-          options: {
-            compilerOptions: {
-              noEmit: false
+    node: {
+        __dirname: false,
+    },
+    target: "electron-main",
+    entry: "./src/electron/main.ts",
+    module: {
+        rules: [
+            {
+                test: /\.ts?$/,
+                use: [{
+                    loader: "ts-loader",
+                    options: {
+                        compilerOptions: {
+                            noEmit: false
+                        }
+                    }
+                }],
+                exclude: /node_modules/
             }
-          }
-        }],
-        exclude: /node_modules/
-      }
-    ]
-  },
-  resolve: {
-    extensions: [".ts", ".js"]
-  },
-  output: {
-    filename: "main.js",
-    path: path.resolve(__dirname, "../build")
-  }
+        ]
+    },
+    resolve: {
+        extensions: [".ts", ".js"]
+    },
+    output: {
+        filename: "main.js",
+        path: path.resolve(__dirname, "../build")
+    }
 };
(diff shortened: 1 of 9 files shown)


config/webpack.common.js
```diff
@@ -1,29 +1,32 @@
...
-  module: {
-    rules: [
-      {
-        test: /\.ts?$/,
-        use: [{
-          loader: "ts-loader",
-          options: {
-            compilerOptions: {
-              noEmit: false
+    node: {
```
wbreza: Part 1 - webpack be default mocks `__dirname`.  We don't want this.

src/App.tsx
```diff
@@ -1,11 +1,11 @@
 import React, { Fragment } from "react";
 import { connect } from "react-redux";
-import { BrowserRouter as Router } from "react-router-dom";
+import { HashRouter as Router } from "react-router-dom";
```
wbreza: Part 2 - React app should use Hash router where all routing info is after the `#` in the url.

wbreza: We'll need to create some switch here between react/browser mode.

src/electron/main.ts
```diff
@@ -8,28 +8,21 @@ let mainWindow: BrowserWindow;
...
 function createWindow() {
-    // and load the index.html of the app.
-
     const windowOptions: BrowserWindowConstructorOptions = {
         width: 1024,
         height: 768,
     };
-    // Create the browser window.
 
+    const staticUrl = process.env.ELECTRON_START_URL || `file:///${__dirname}/index.html`;
```
wbreza: Part 3 - With other changes in place we can now simplify our startup to always use `loadURL(...)`

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -103,9 +103,11 @@ export default class HomePage extends React.Component<IHomepageProps> {
         );
     }
 
-    private createNewProject = () => {
+    private createNewProject = (e: SyntheticEvent) => {
         this.props.actions.closeProject();
         this.props.history.push("/projects/create");
+
+        e.preventDefault();
```
wbreza: Part 4 - This was the main culprit of the issue navigating to the new project screen.  This button/link is an HTML `<a/>` tag that had a `#` in the `href` attribute.  We don't actually want the browser to navigate using this value so we need to call `e.preventDefault()` to prevent the navigation from occurring and just use our `history.push(...)` call.

src/App.tsx
tbarlow12: @wbreza would some of Jordan's work be useful here?

tbarlow12:
Also, is there a way to disable the patch requirement for this commit? Not entirely sure how we test this change...

src/App.tsx
wbreza: Yes, definitely - just didn't want to include it as part of this PR.

wbreza:
@tbarlow12 - Yes i will bypass the patch requirement for this PR.

src/electron/main.ts
elizabethhalper: So if webpack isn't mocking __dirname, where is it set?

wbreza: `__dirname` is a special node variable that gives you the current directory path of the currently executing code.

PR #630, opened 2019-03-01
fix: Updated EditorTagsInput to work with vott-react

tbarlow12:
The tests will be failing because we don't have a good version of `vott-react` published yet, but as soon as we do, I'll update the version here and the tests should work.

UPDATE:

I manually published `vott-react` 0.2.6 to unblock this side of things, but we still need to fix the CI for `vott-react` to publish properly

UPDATE:

`vott-react` 0.2.7 is now published, so I changed the version here as well


tbarlow12:
This PR now includes the following:
- Update to use `vott-react` 0.2.6
- Removal of obsolete `EditorTagsInput` - `TagsInput` now contained directly inside the `EditorFooter`, which holds the new function `getTagSpan` that produces the contents and style of the text inside of each tag
- Display of tags that are locked by using, bold, white font

src/react/components/pages/editorPage/editorFooter.tsx
```diff
@@ -84,22 +76,43 @@ export default class EditorFooter extends React.Component<IEditorFooterProps, IE
...
 
-    private onTagsChanged(tags) {
+    private onTagsChanged = (tags) => {
         this.setState({
             tags,
         }, () => this.props.onTagsChanged(this.state));
     }
+
+    /**
+     * Shows the of the tag in the span of the first 10 tags
```
mydiemho: missing a word in description

src/react/components/pages/editorPage/editorFooter.tsx
```diff
@@ -84,22 +76,43 @@ export default class EditorFooter extends React.Component<IEditorFooterProps, IE
...
+     * @param name Name of tag
+     * @param index Index of tag
+     */
+    private getTagSpan = (name: string, index: number) => {
+        let className = "tag-span";
+        let displayName = name;
+        if (index < 10) {
+            const displayIndex = (index === 9) ? 0 : index + 1;
```
mydiemho: what is the reasoning behind this index math?

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -132,6 +132,12 @@
     }
 }
 
+.locked-tag {
```
mydiemho: Could you add a screenshot to show what this change look like?

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -174,8 +174,8 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
                     </div>
                     <div>
                         <EditorFooter
-                            displayHotKeys={true}
```
mydiemho: why are we replacing this with lockedTags/

tbarlow12: because `displayHotKeys` is not longer needed and `lockedTags` now is

src/react/components/pages/editorPage/editorFooter.tsx
tbarlow12: The display index is not the same as the index in the array. We want the 10th tag to be 0 and every other to be `index + 1`

src/react/components/pages/editorPage/editorFooter.tsx
tbarlow12: Good catch

PR #631, opened 2019-03-01
feature: Enhanced error message for unknown errors (not in prod mode)

tbarlow12:


src/registerToolbar.ts
```diff
@@ -89,6 +89,7 @@ export default function registerToolbar() {
         icon: "fa-trash-alt",
         group: ToolbarItemGroup.Regions,
         type: ToolbarItemType.Action,
+        accelerators: ["Ctrl+Delete"],
```
mydiemho: should put this in a different PR, I prefer not mixing code

src/react/components/common/errorHandler/errorHandler.tsx
```diff
@@ -79,7 +79,6 @@ export class ErrorHandler extends React.Component<IErrorHandlerProps> {
     private handleError(error: string | Error | AppError) {
         console.log(error);
```
mydiemho: we should also remove this left over code

src/react/components/common/errorHandler/errorHandler.tsx
```diff
@@ -103,29 +102,40 @@ export class ErrorHandler extends React.Component<IErrorHandlerProps> {
                 message: reason.message,
                 title: reason.name,
             };
+        } else {
+            appError = {
```
mydiemho: I think we can combine this with the previous case.  sometimes it happens that type Error has no message or name

src/registerToolbar.ts
tbarlow12: Cool. Will do

src/react/components/common/errorHandler/errorHandler.tsx
```diff
@@ -103,29 +102,40 @@ export class ErrorHandler extends React.Component<IErrorHandlerProps> {
...
+                message: this.getUnknownErrorMessage(error),
+            };
         }
+        this.props.onError(appError);
+    }
 
-        if (!appError) {
-            appError = new AppError(ErrorCode.Unknown, "Unknown Error occurred");
+    private getUnknownErrorMessage(e) {
+        if (Env.get() !== "production") {
```
wbreza: I'm assuming when we do a prod build `create-react-app` sets node_env to "production" ?

tbarlow12: That's what I understood. I could be wrong, but I'll look into this further

tbarlow12: Verified this with a release version.

tbarlow12:
Closing in favor of [PR against dev branch](https://github.com/Microsoft/VoTT/pull/639)

PR #635, opened 2019-03-04
fix: Tagging new region only applies to one region [AB#17176]

tbarlow12:
- Canvas Helpers now returns clones of transformed tag arrays
- Canvas now allows CanvasTools to manage selected regions (removed from Canvas state)
- Fixing tag tests from editor page


src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -413,14 +424,6 @@ describe("Editor Page Component", () => {
...
         });
 
-        function dispatchKeyEvent(key: string, eventType: KeyEventType = KeyEventType.KeyDown) {
-            window.dispatchEvent(new KeyboardEvent(
-                eventType, {
-                    key,
-                },
-            ));
-        }
-
```
elizabethhalper: Why are you getting rid of this function?

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -498,6 +501,27 @@ describe("Editor Page Component", () => {
             dispatchKeyEvent("Ctrl+e");
             expect(clearRegions).toBeCalled();
         });
```
elizabethhalper: Make sure you are up to date with the current dev. The Ctrl+e is related to the export provider and I don't know if this test is valid anymore.

tbarlow12: Good call. I'll make the change

tbarlow12: Good call. I'll make the change

src/react/components/pages/editorPage/editorPage.test.tsx
tbarlow12: I think I moved it somewhere else... I'll look into it

PR #636, opened 2019-03-05
(WIP)fix: Export ALL assets VS Visited/Tagged [#AB17365]

JacopoMangiavacchi:
- Add allAssets list to project and persist on project metadata
- Update state.project from props.project in editorPage
- Use state.project to save and export project
- Differentiate source list when exporting All VS Visited/Tagged
- Re-add exportProject to toolbar
- Update different test

FIX bug #AB17365

### DIFF
--- a/src/providers/export/tensorFlowRecords.test.ts
+++ b/src/providers/export/tensorFlowRecords.test.ts
@@ -14,6 +14,7 @@ jest.mock("../storage/localFileSystemProxy");
 import { LocalFileSystemProxy } from "../storage/localFileSystemProxy";
 import registerMixins from "../../registerMixins";
 import { appInfo } from "../../common/appInfo";
+import { asin } from "snapsvg";
 
 registerMixins();
 
@@ -25,6 +26,7 @@ describe("TFRecords Json Export Provider", () => {
         "asset-3": MockFactory.createTestAsset("3", AssetState.Visited),
         "asset-4": MockFactory.createTestAsset("4", AssetState.NotVisited),
     };
+    baseTestProject.allAssets = _.values(baseTestProject.assets);
     baseTestProject.sourceConnection = MockFactory.createTestConnection("test", "localFileSystemProxy");
     baseTestProject.targetConnection = MockFactory.createTestConnection("test", "localFileSystemProxy");
 
--- a/src/react/components/pages/editorPage/editorPage.tsx
+++ b/src/react/components/pages/editorPage/editorPage.tsx
@@ -175,7 +175,7 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
                     <div>
                         <EditorFooter
                             displayHotKeys={true}
-                            tags={this.state.project.tags}
+                            tags={this.props.project.tags}
                             onTagsChanged={this.onFooterChange}
                             onTagClicked={this.onTagClicked}
                             onCtrlTagClicked={this.onCtrlTagClicked}
--- a/src/react/components/pages/editorPage/editorPage.tsx
+++ b/src/react/components/pages/editorPage/editorPage.tsx
@@ -113,9 +113,9 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         // Navigating directly to the page via URL (ie, http://vott/projects/a1b2c3dEf/edit) sets the default state
         // before props has been set, this updates the project and additional settings to be valid once props are
         // retrieved.
-        if (!this.state.project && this.state.project) {
+        if (this.state.project !== this.props.project) {
             this.setState({
-                project: this.state.project,
+                project: this.props.project,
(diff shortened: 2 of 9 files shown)


JacopoMangiavacchi:
@wbreza main bug was a regression due to some recent PR removing the the list of all assets (even non visited/tagged) from the editorPage.props.asset list.  Phil and I discussed best approach and we agreed to add a new list for all assets in the IProject interface.

@PIC123 There are still a couple of tests failing.  One is the same as on @elizabethhalper recent PR #634 and it's about the Ctrl+e that is not calling the clearRegion handler.  Second one is a little bit tricky and is on a test forcing to load the editorPage with an empty project prop.  I can't understand why the didComponentUpdate lifecycle that seams to work pretty well at runtime is having this issue returning after the flushUI call a state.project different than prop.project.

src/providers/export/tensorFlowRecords.test.ts
```diff
@@ -14,6 +14,7 @@ jest.mock("../storage/localFileSystemProxy");
 import { LocalFileSystemProxy } from "../storage/localFileSystemProxy";
 import registerMixins from "../../registerMixins";
 import { appInfo } from "../../common/appInfo";
+import { asin } from "snapsvg";
```
wbreza: Dont' think you need this here.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -97,7 +97,7 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
 
     public async componentDidMount() {
         const projectId = this.props.match.params["projectId"];
-        if (this.props.project) {
+        if (this.state.project) {
```
wbreza: We shouldn't need `state` project here.

wbreza: We should deprecate `project` from the component state since we are not editing without saving the project.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -106,17 +106,17 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
     }
 
     public async componentDidUpdate() {
-        if (this.props.project && this.state.assets.length === 0) {
+        if (this.state.project && this.state.assets.length === 0) {
```
wbreza: Convert all these back to `this.props.project`

mydiemho:
@JacopoMangiavacchi  I have remove `All Assets` from the dropdown and change default to `Visited` until you have a fix.  When your change is ready, please also add back `All Assets` 

https://github.com/Microsoft/VoTT/pull/646

src/providers/export/tensorFlowRecords.test.ts
JacopoMangiavacchi: Sorry, I really have no idea how this line ended up here.  I'll definitely remove this.

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: Completely agree.  If we don't want to alter the project metadata adding the "all" assets list then definitely it should be better to completely remove project from the state as it will return to be a read only prop.

JacopoMangiavacchi:
@wbreza the reason why we decided to add the "all" asset list to the project was to avoid to browse the source provider asset list twice.  We wondered that in case this is a network provider there's an extra cost for doing that and we wanted to avoid this as this list had already been browsed in the editorPage on project loading so we though it could be nice to cache this directly in the project.  Eventually we can cache it on a different redux state but but I still think there are several reason why it should be good to add this to the project.  

Btw in an initial implementation this was just the list of IDs but I figured it out later that for handling TFRecord input I needed also the asset file extension so I ended up to store the entire asset object in this list.  

Anyway I think that having this list in the project file could be particularly useful in the future in case we want to allows to support importing assets from multiple source providers or even multiple configurations of the same asset provider (aka different query parameter for the Bing image search).  

Are you sure you still don't like to add this list to the project ? ;-)

wbreza:
Thanks for the info @JacopoMangiavacchi, but i would still rather not include it in the project.  I agree with you that when people export their project they will likely either use visited or tagged assets.  

If you think about large projects where there are thousands of assets it is just unnecessary overhead to keep these in the project file.  With regard on online searches like bing this is only Page 1 of data... Right now we don't even consider Page 2 and beyond.  Network requests are low overhead and we do this every time the user navigates to the editor page.  Moving forward have multiple source providers, random search, etc I can imagine the project file getting very bloated very quickly if we store every asset we ever retrieved in the project file.

JacopoMangiavacchi:
@wbreza about the project file size we could probably just simply map the IAsset list and only save asset ID and Path in this new list.  Anyway fair enough, let's just re-browse from the provider when exporting.  Are you fixing this on a different PR ?  I can work on this later today probably or tomorrow in the afternoon eventually.

PR #640, opened 2019-03-05
fix: Don't allow invalid pasting [AB#17353]

tbarlow12:
- Still need to implement tests for most of this
- Scales pasted regions appropriately
- Pastes regions that don't fit to top left and cascades down and to the right
- Throws error if region is too big to be pasted in asset


wbreza:
Looking good after tests pass.

PR #641, opened 2019-03-05
fix: save tagged video frames as "jpg" files

elizabethhalper:
Video frames were not correctly exported when using Pascal VOC. Images were generated but did't have "jpg" file extension; it is expected that all video frames are exported to JPG images when project is exported

### DIFF
--- a/src/common/htmlFileReader.test.ts
+++ b/src/common/htmlFileReader.test.ts
@@ -194,22 +194,36 @@ describe("Html File Reader", () => {
     });
 
     describe("Extracting video frames", () => {
-        const videoAsset = MockFactory.createVideoTestAsset("VideoTestAsset", AssetState.Tagged);
-        const videoFrame = MockFactory.createChildVideoAsset(videoAsset, 123.456);
-
-        beforeEach(() => {
-            assetTestCache.set(videoFrame.parent.path, videoFrame);
-        });
-
         it("Gets a blob for the requested video frame", async () => {
+            const videoAsset = MockFactory.createVideoTestAsset("VideoTestAsset-1", AssetState.Tagged);
+            const videoFrame = MockFactory.createChildVideoAsset(videoAsset, 123.456);
+            assetTestCache.set(videoFrame.parent.path, videoFrame);
+
             const blob = await HtmlFileReader.getAssetFrameImage(videoFrame);
             expect(blob).not.toBeNull();
             expect(blob).toBeInstanceOf(Blob);
         });
 
-        it("Appends jpg file extension on specified asset", async () => {
+        it("Appends jpg file extension on specified video frame asset", async () => {
+            const videoAsset = MockFactory.createVideoTestAsset("VideoTestAsset-2", AssetState.Tagged);
+            const videoFrame = MockFactory.createChildVideoAsset(videoAsset, 456.789);
+            assetTestCache.set(videoFrame.parent.path, videoFrame);
+
+            expect(videoFrame.name.endsWith(".jpg")).toBe(false);
             await HtmlFileReader.getAssetFrameImage(videoFrame);
-            expect(videoFrame.name.endsWith(".jpg"));
+            expect(videoFrame.name.endsWith(".jpg")).toBe(true);
+        });
+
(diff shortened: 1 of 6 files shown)


src/providers/export/tensorFlowPascalVOC.ts
```diff
@@ -75,7 +75,7 @@ export class TFPascalVOCJsonExportProvider extends ExportProvider<ITFPascalVOCJs
     }
 
     private async exportSingleImage(jpegImagesFolderName: string, element: IAssetMetadata): Promise<void> {
-        const imageFileName = `${jpegImagesFolderName}/${element.asset.name}`;
+        const imageFileName = `${jpegImagesFolderName}/${element.asset.name.replace(/\.[^/.]+$/, ".jpg")}`;
```
wbreza: Agreed - need some more tests.  How can we validate that this happens consistently across providers?

elizabethhalper: @wbreza I agree, but I'm not sure how to do this given the tests we have are already "testing" that assets are exported to "jpg" images. Let's have a discussion offline as to how to refactor them to test the correct thing.

src/common/utils.test.ts
```diff
@@ -50,5 +50,12 @@ describe("Helper functions", () => {
 
             expect(decryptedProject).toEqual(testProject);
         });
+
+        it("Replaces an extension with new extension", () => {
+            expect(replaceExtension("filename.badext", ".jpg")).toEqual("filename.jpg");
+            expect(replaceExtension("filename.this.is.bad", ".jpg")).toEqual("filename.this.is.jpg");
+            expect(replaceExtension("filename", ".jpg")).toEqual("filename");
```
mydiemho: if a file has no extension, should we add an extension?

src/common/htmlFileReader.test.ts
```diff
@@ -3,8 +3,26 @@ import HtmlFileReader from "./htmlFileReader";
...
 import MockFactory from "./mockFactory";
+import { AssetState, IAsset } from "../models/applicationState";
 
 describe("Html File Reader", () => {
+    const assetTestCache = new Map<string, IAsset>();
+
+    beforeEach(() => {
+        assetTestCache.clear();
+
+        document.createElement = jest.fn((elementType) => {
```
wbreza: Refactored html element mocks.

src/common/htmlFileReader.ts
```diff
@@ -104,12 +104,21 @@ export default class HtmlFileReader {
...
+     * Extracts the specified image frame from a video asset
+     * @param asset The asset video frame to retrieve from the parent video
+     */
+    public static async getAssetFrameImage(asset: IAsset): Promise<Blob> {
         return new Promise((resolve, reject) => {
             const cachingEnabled = false;
             let refresh = !cachingEnabled;
             let video: HTMLVideoElement = this.videoAssetFiles[asset.parent.name];
 
+            // Ensure the asset name includes jpg file extension
```
wbreza: Moving fix into HtmlFileReader to easily apply to all export providers.

src/providers/export/vottJson.ts
```diff
@@ -32,7 +32,7 @@ export class VottJsonExportProvider extends ExportProvider<IVottJsonExportOption
         await results.forEachAsync(async (assetMetadata) => {
             return new Promise<void>(async (resolve) => {
                 const blob = await HtmlFileReader.getAssetBlob(assetMetadata.asset);
-                const assetFilePath = `vott-json-export/${assetMetadata.asset.id}.jpg`;
+                const assetFilePath = `vott-json-export/${assetMetadata.asset.name}`;
```
wbreza: Normalized VoTT json to follow same convensions as other export providers that same image binaries.

src/providers/export/tensorFlowPascalVOC.ts
wbreza: Moved fix into HtmlFileReader and added tests.

src/common/utils.test.ts
wbreza: Moved fix into `HtmlFileReader` and added tests.  Will always add "jpg" extension if it does not already exist.

PR #642, opened 2019-03-05
fix: set default export format in project

mydiemho:
export wasn't doing anything because there wasn't a default format.

AB#17437

### DIFF
--- a/src/models/applicationState.ts
+++ b/src/models/applicationState.ts
@@ -44,6 +44,7 @@ export enum ErrorCode {
     ProjectUploadError = "projectUploadError",
     ProjectDeleteError = "projectDeleteError",
     SecurityTokenNotFound = "securityTokenNotFound",
+    ExportFormatNotFound = "exportFormatNotFound",
     CanvasError = "canvasError",
     V1ImportError = "v1ImportError",
 }
--- a/src/redux/actions/projectActions.test.ts
+++ b/src/redux/actions/projectActions.test.ts
@@ -10,7 +10,7 @@ import ProjectService from "../../services/projectService";
 jest.mock("../../services/assetService");
 import { AssetService } from "../../services/assetService";
 import { ExportProviderFactory } from "../../providers/export/exportProviderFactory";
-import { IExportProvider } from "../../providers/export/exportProvider";
+import { ExportAssetState, IExportProvider } from "../../providers/export/exportProvider";
 import { IApplicationState } from "../../models/applicationState";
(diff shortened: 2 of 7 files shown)


src/redux/actions/projectActions.ts
```diff
@@ -1,26 +1,38 @@
...
+
     closeProject(): void;
+
     exportProject(project: IProject): Promise<void> | Promise<IExportResults>;
+
     loadAssets(project: IProject): Promise<IAsset[]>;
+
     loadAssetMetadata(project: IProject, asset: IAsset): Promise<IAssetMetadata>;
+
     saveAssetMetadata(project: IProject, assetMetadata: IAssetMetadata): Promise<IAssetMetadata>;
```
elizabethhalper: nit: Any reason for adding the extra spaces? I think when exporting an interface it is more readable and consistent with our other formatting to leave them out.

src/redux/actions/projectActions.test.ts
```diff
@@ -76,14 +76,41 @@ describe("Project Redux Actions", () => {
...
-
         const result = await projectActions.saveProject(project)(store.dispatch, store.getState);
 
         expect(result.version).toEqual(appInfo.version);
     });
 
+    it("Save Project action on new project correctly add default export format", async () => {
+        projectServiceMock.prototype.save = jest.fn((project) => Promise.resolve(project));
+
+        const skeletonProject = MockFactory.createTestProject("TestProject");
```
elizabethhalper: nit: I would call this either "project" or "testProject" like we have been doing in the rest of the tests

mydiemho: project is below, this one is use to create the test project

src/redux/actions/projectActions.ts
mydiemho: oh autoformat on webstorm did that, I can undo it

src/models/applicationState.ts
```diff
@@ -44,6 +44,7 @@ export enum ErrorCode {
     ProjectUploadError = "projectUploadError",
     ProjectDeleteError = "projectDeleteError",
     SecurityTokenNotFound = "securityTokenNotFound",
+    ExportFormatNotFound = "exportFormatNotFound",
```
wbreza: When specifying a new error code you will also need to setup the localized strings for the error code.

mydiemho: fixed

PR #645, opened 2019-03-06
fix: remove grey and white from tagColors

elizabethhalper:
Tags with grey and white colors associated with them are not recognized by canvas tools and therefore are rendered in the editorPage as red. Eliminating these colors to fix the bug.


src/react/components/common/tagColors.json
```diff
@@ -1,6 +1,4 @@
 [
-    "#FFFFFF",
```
tbarlow12: I think we actually might be able to get rid of this file entirely now. I had forgotten that it was still there

tbarlow12: If not, we should update to use the object from `tagColors.ts`

elizabethhalper: I'll try getting rid of it and see if I run into any issues

wbreza: @JacopoMangiavacchi is using the colors somewhere in VoTT. I believe within the TF records

package.json
```diff
@@ -43,7 +43,7 @@
         "shortid": "^2.2.14",
         "video-react": "^0.13.2",
         "vott-ct": "^2.1.19",
-        "vott-react": "^0.2.8"
+        "vott-react": "^0.2.9"
```
wbreza: Looks like package-lock is causing a failure... Not sure how you updated this.  Was it manual? or via `npm install`?  You will have to run a real `npm install` for the package lock to get updated correctly.

src/react/components/common/tagColors.json
JacopoMangiavacchi: Yes I'm using this when importing TFRecords.  In that scenario we are importing potentially new tag definitions that have to be added to IProject.ITags list.  Basically I need to associate a new color from that list to the new ITag object.

PR #648, opened 2019-03-06
fix: Ignore error from react drag n drop [AB#17466]

tbarlow12:
ReactDnD currently throws an error when tags are dragged. The operation still performs successfully, but throws the error anyway. This catches that specific error and ignores it


src/react/components/common/errorHandler/errorHandler.tsx
```diff
@@ -77,6 +77,14 @@ export class ErrorHandler extends React.Component<IErrorHandlerProps> {
      * @param error The error to handle
      */
     private handleError(error: string | Error | AppError) {
+        // This is a special case where we don't need to throw an
```
wbreza: Is there no way for us to catch this within the tags input component?  Fix the underlying cause?

tbarlow12: It's thrown by the ReactDnD inside the tags input component we're using... My fix is the new tags input component. This is just the patch for the time being

PR #649, opened 2019-03-07
fix: Resolves issue exporting all assets

wbreza:
Resolves issue where exporting all assets doesn't include non-visited assets.  Non-visited assets are no longer part of the project file and requires merging in assets from the source connection.

Resolves [AB#17365](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17365)


src/providers/export/exportProvider.ts
```diff
@@ -72,7 +73,18 @@ export abstract class ExportProvider<TOptions> implements IExportProvider {
     public async getAssetsForExport(): Promise<IAssetMetadata[]> {
         let predicate: (asset: IAsset) => boolean = null;
 
-        // @ts-ignore
+        const getProjectAssets = () => Promise.resolve(_.values(this.project.assets));
```
wbreza: This is the core of the fix.  Defaults to projectAssets when options is set to `Visited` or `Tagged`, otherwise merges in assets from the source connection when option is set to `All`.

src/providers/export/tensorFlowPascalVOC.ts
```diff
@@ -1,19 +1,11 @@
...
-import { ExportProvider, ExportAssetState } from "./exportProvider";
-import { IProject, IAssetMetadata, RegionType, ITag } from "../../models/applicationState";
+import { ExportProvider } from "./exportProvider";
+import { IProject, IAssetMetadata, RegionType, ITag, IExportProviderOptions } from "../../models/applicationState";
 import Guard from "../../common/guard";
 import HtmlFileReader from "../../common/htmlFileReader";
 import { itemTemplate, annotationTemplate, objectTemplate } from "./tensorFlowPascalVOC/tensorFlowPascalVOCTemplates";
 import { interpolate } from "../../common/strings";
 
-/**
```
wbreza: Removed types that were same as default export provider options.

src/providers/export/exportProvider.ts
```diff
@@ -72,7 +73,18 @@ export abstract class ExportProvider<TOptions> implements IExportProvider {
...
+        const getAllAssets = async () => {
+            const projectAssets = await getProjectAssets();
+
+            return _(projectAssets)
```
mydiemho: what is the `_` in front does?

src/providers/export/exportProvider.ts
```diff
@@ -82,11 +94,12 @@ export abstract class ExportProvider<TOptions> implements IExportProvider {
...
             case ExportAssetState.All:
             default:
+                getAssetsFunc = getAllAssets;
                 predicate = () => true;
                 break;
         }
 
-        return await _.values(this.project.assets)
+        return (await getAssetsFunc())
             .filter((asset) => asset.type !== AssetType.Video)
```
mydiemho: so we are still ignoring videos in "All asset" export?

src/providers/export/tensorFlowPascalVOC.test.ts
```diff
@@ -28,6 +32,7 @@ function _base64ToArrayBuffer(base64: string) {
 }
 
 describe("TFPascalVOC Json Export Provider", () => {
+    const testAssets = MockFactory.createTestAssets(10, 1);
     const baseTestProject = MockFactory.createTestProject("Test Project");
     baseTestProject.assets = {
         "asset-1": MockFactory.createTestAsset("1", AssetState.Tagged),
```
mydiemho: should we replace these with the `testAssets` created above?

src/providers/export/exportProvider.ts
wbreza: `_` is [lodash](https://lodash.com)

src/providers/export/exportProvider.ts
wbreza: we ignore `AssetType.Video` for all exports.  All `Visited` or `Tagged` Video Frames are exported but we don't export the container Video asset itself.  Hope that clears up any confusion.

src/providers/export/tensorFlowPascalVOC.test.ts
wbreza: No - these testAssets are equivalent to the `NonVisted` assets that are not part of the project file.  The ones defined elsewhere in these tests are those specifically included in the project file as `Visited` or `Tagged`.

src/providers/export/tensorFlowPascalVOC.test.ts
```diff
@@ -118,14 +131,14 @@ describe("TFPascalVOC Json Export Provider", () => {
...
-            expect(writeBinaryCalls.length).toEqual(4);
+            expect(writeBinaryCalls.length).toEqual(testAssets.length);
             expect(writeBinaryCalls[0][0].endsWith("/JPEGImages/Asset 1.jpg")).toEqual(true);
             expect(writeBinaryCalls[1][0].endsWith("/JPEGImages/Asset 2.jpg")).toEqual(true);
             expect(writeBinaryCalls[2][0].endsWith("/JPEGImages/Asset 3.jpg")).toEqual(true);
             expect(writeBinaryCalls[3][0].endsWith("/JPEGImages/Asset 4.jpg")).toEqual(true);
 
             const writeTextFileCalls = storageProviderMock.mock.instances[0].writeText.mock.calls as any[];
-            expect(writeTextFileCalls.length).toEqual(11);
+            expect(writeTextFileCalls.length).toEqual(17);
```
tbarlow12: Any way to calculate this number from other variables? Or document why it's 17? Just to make the test a little less fragile if things change

wbreza: Totally agree with you, was able to figure out the computation.

JacopoMangiavacchi: We could remove this line.  That is the total number of text files exported that for pascal voc is something like 1 pbtxt + 1 * tag * 2 (train/validation) + 1 per asset.  But we test the presence of these specific files few lines below in same test so we could remove it.

PR #650, opened 2019-03-07
fix: lock tag not working on mac

mydiemho:
using latest vott-react that includes the fix in the `tagInputs` component

AB#17483

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6999,7 +6999,8 @@
                 },
                 "ansi-regex": {
                     "version": "2.1.1",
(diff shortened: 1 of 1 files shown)


mydiemho:
> Shouldn't package.json be updated to reference the new version?

someone has already bump the version at some point https://github.com/Microsoft/VoTT/blob/dev/package.json#L46

mydiemho:
Previous PR #645 already bump vott-react which included this fix, so will be closing PR

Issue #573, opened 2019-02-13
V2 Preview - Quick Square fuction not in V2


JCAAUSFRA:
In previous versions there was a function where you could scroll and create a square around the desired region. you could then adjust the size of the square by pressing ctrl and scrolling. This was very useful.

wbreza:
Thanks for the feedback @JCAAUSFRA  - by the end of our v2 development cycle we plan to achieve parity with the most used/wanted features that currently exist in v1

wbreza:
This feature will be complete with our Preview 3 release.

wbreza:
Resolved with #651

wbreza:
Feature is available within v2 Preview 3

PR #651, opened 2019-03-07
feat: Enable copy rectangle functionality in editor

PIC123:
This adds the copy rectangle functionality of canvas tools to the VOTT editor. When a region is selected on the canvas, the user can select the copyRect button to switch modes and create regions, using the shape of the previously selected region as a template.

*New PR because of issues with branch history

Resolves [[AB#17074](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17074)]


src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -59,7 +60,7 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
         this.editor.onRegionMoveEnd = this.onRegionMoveEnd;
         this.editor.onRegionDelete = this.onRegionDelete;
         this.editor.onRegionSelected = this.onRegionSelected;
-        this.editor.AS.setSelectionMode(this.props.selectionMode, null);
+        this.editor.AS.setSelectionMode({mode: this.props.selectionMode});
```
wbreza: Fix whitespace.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -80,7 +81,19 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
         }
 
         if (this.props.selectionMode !== prevProps.selectionMode) {
-            this.editor.AS.setSelectionMode(this.props.selectionMode, null);
+            let  options = null;
+            if (this.props.selectionMode === SelectionMode.COPYRECT) {
+                const selectedRegion = this.getSelectedRegions();
```
wbreza: Doesn't this return an array of regions?  If no selected regions does it return null or array of zero?  Please confirm and update logic as needed.

Also - I feel that the `componentDidUpdate` is not the place to specify the template options.  I would rather see the template selection happen on region selected or on draw of the region

src/registerToolbar.ts
```diff
@@ -57,6 +57,15 @@ export default function registerToolbar() {
         accelerators: ["p", "P"],
     });
 
+    ToolbarItemFactory.register({
+        name: ToolbarItemName.CopyRectangle,
+        tooltip: "Copy Rectangle",
```
wbreza: Nit: Localize strings.

src/registerToolbar.ts
```diff
@@ -57,6 +57,15 @@ export default function registerToolbar() {
         accelerators: ["p", "P"],
     });
 
+    ToolbarItemFactory.register({
```
wbreza: Let's also add a test to very the copy rect button renders in the toolbar since we have been bitten in the past where buttons were accidentally removed.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -42,6 +43,8 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
 
     public editor: Editor;
 
+    public template: Rect = new Rect(20, 20);
```
wbreza: This shouldn't need to be public.

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -325,13 +331,15 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
     /**
      * Method called when deleting a region from the editor
-     * @param {string} id the id of the deleted region
+     * @param {string} id the id of the selected region
      * @param {boolean} multiselect boolean whether region was selected with multiselection
      * @returns {void}
      */
     private onRegionSelected = (id: string, multiselect: boolean) => {
+        const selectedRegions = this.getSelectedRegions();
+        // const selectedRegion = selectedRegions.find((region) => region.id === id);
```
wbreza: Seems like you were on the write track with this commented out code... if you select a new region do we want to set that as the template??

src/registerToolbar.ts
wbreza: Still not seeing this.

src/common/localization/es-cl.ts
```diff
@@ -166,10 +166,17 @@ export const spanish: IAppStrings = {
...
-            exportProject: "Exprtar Proyecto",
+            drawRectangle: "Dibujar Rectángulo (R)",
+            drawPolygon: "Dibujar Polígono (P)",
+            copyRectangle: "Copia rectángulo (Ctrl + W)",
+            copy: "Copiar regiones (Ctrl + C)",
+            cut: "Cortar regiones (Ctrl + X)",
+            paste: "Pegar regiones (Ctrl + V)",
+            removeAllRegions: "Eliminar Todas Las Regiones (Ctrl + Delete)",
+            previousAsset: "Activo anterior (W)",
+            nextAsset: "Siguiente activo (S)",
```
tbarlow12: Activo siguiente

wbreza:
Resolves #573

PR #654, opened 2019-03-08
fix: Display asset load error when an asset fails to load

PIC123:
Fixes bug where app would throw errors when assets would not load and did not fail gracefully. Now asset preview in sidebar and canvas area is replaced with an appropriate error message if the asset cannot load.

image

Resolves [[AB#17207](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17207)]

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -1,5 +1,9 @@
 @import './assets/sass/theme.scss';
 
+body {
+    user-select: none;
+}
+
 #root {
     height: 100vh;
     overflow: hidden;
--- a/src/models/applicationState.ts
+++ b/src/models/applicationState.ts
@@ -229,6 +229,7 @@ export interface IAsset {
     format?: string;
     timestamp?: number;
     parent?: IAsset;
+    hasError?: boolean;
 }
 
 /**
--- a/src/models/applicationState.ts
+++ b/src/models/applicationState.ts
@@ -229,7 +229,6 @@ export interface IAsset {
     format?: string;
     timestamp?: number;
     parent?: IAsset;
-    hasError?: boolean;
 }
 
 /**
--- a/src/react/components/common/assetPreview/assetPreview.tsx
+++ b/src/react/components/common/assetPreview/assetPreview.tsx
@@ -25,6 +25,7 @@ export interface IAssetProps {
     onDeactivated?: (contentSource: ContentSource) => void;
     /** Event handler that fires when a child asset is selected (ex. Paused on a video frame) */
     onChildAssetSelected?: (asset: IAsset) => void;
+    onAssetError?: (event: React.SyntheticEvent<HTMLImageElement>) => void;
 }
 
 /**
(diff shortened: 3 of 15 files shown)


src/react/components/common/assetPreview/assetPreview.tsx
```diff
@@ -85,38 +86,40 @@ export class AssetPreview extends React.Component<IAssetPreviewProps, IAssetPrev
...
-                    {asset.type === AssetType.TFRecord &&
-                        <TFRecordAsset asset={asset}
-                            onLoaded={this.onAssetLoad}
-                            onActivated={this.props.onActivated}
-                            onDeactivated={this.props.onDeactivated} />
-                    }
-                    {asset.type === AssetType.Unknown &&
-                        <div className="asset-error">{strings.editorPage.assetError}</div>
-                    }
+                    {!asset.hasError ?
```
wbreza: Let's invert this expression, if there is an asset error then show the new experience and return.  Consider breaking down the render function into smaller render methods.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -154,7 +156,7 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
                             onToolbarItemSelected={this.onToolbarItemSelected} />
                     </div>
                     <div className="editor-page-content-body">
```
wbreza: Since the `AssetPreview` experience already displays an error you shouldn't need to do anything here directly in the editor page.

src/models/applicationState.ts
```diff
@@ -229,6 +229,7 @@ export interface IAsset {
     format?: string;
     timestamp?: number;
     parent?: IAsset;
+    hasError?: boolean;
```
wbreza: I don't think we need to introduce a new modal property here that is persisted since the error may be transient.  If we REALLY wanted to store something then I would recommend we extend the `AssetState` to include a new enum for `Error`.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -368,6 +373,12 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         }
     }
 
+    private onAssetError = () => {
```
wbreza: What is the goal here? To persisted that the asset has an error?  Seems unnecessary.

src/react/components/pages/editorPage/editorSideBar.tsx
```diff
@@ -13,14 +13,14 @@ export interface IEditorSideBarProps {
     assets: IAsset[];
     onAssetSelected: (asset: IAsset) => void;
     selectedAsset?: IAsset;
+    onAssetError?: () => void;
```
wbreza: I don't feel that an individual asset error warrants bubbling this up - Let is control its display rendering internally.

src/react/components/pages/editorPage/editorSideBar.tsx
```diff
@@ -100,14 +98,15 @@ export default class EditorSideBar extends React.Component<IEditorSideBarProps,
 
     private rowRenderer({ key, index, style }) {
         const asset = this.props.assets[index];
-        const { selectedAsset } = this.state;
+        const selectedAsset = this.props.selectedAsset;
```
wbreza: Thanks for clearing up redundant state where we can use the props directly 👍

src/index.scss
```diff
@@ -1,5 +1,9 @@
 @import './assets/sass/theme.scss';
 
+body {
```
wbreza: Fixes issue where app text/images are not selecatable when click-drag && Ctrl+A to select regions

wbreza:
FYI - CodeCove comment on top of PR did not update.  Current coverage is ~91%

PR #656, opened 2019-03-08
ci: script to deploy VoTT to Azure

pjlittle:
Script to push VoTT single page app to Azure as a static Web site.
Display version/commit SHA in multiple places to be able to easily
determine version.

[AB#17484](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17484)

Preview deployment available here: https://vottdev.z5.web.core.windows.net

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -19,13 +19,14 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>React App</title>
+    <title>Visual Object Tagging Tool (VoTT) v%REACT_APP_VERSION%</title>
   </head>
   <body>
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
     <div id="root"></div>
+    <div id="appInfo">VoTT v%REACT_APP_VERSION%, commit=%REACT_APP_COMMIT_SHA%</div>
     <!--
       This HTML file is a template.
       If you open it directly in the browser, you will see an empty page.
--- a/src/index.scss
+++ b/src/index.scss
@@ -5,6 +5,10 @@
     overflow: hidden;
 }
 
+#appInfo {
+    display: none;
+}
+
 input[type=file] {
     display: none;
 }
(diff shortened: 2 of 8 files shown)


src/react/components/pages/appSettings/appSettingsPage.tsx
```diff
@@ -64,6 +64,9 @@ export default class AppSettingsPage extends React.Component<IAppSettingsProps>
                     <div className="my-3">
                         <p>{`${strings.appSettings.version.description} ${appInfo.version}`}</p>
                     </div>
+                    <div className="my-3">
+                        <p>{`${strings.appSettings.commit}: `} {process.env.REACT_APP_COMMIT_SHA}</p>
```
pjlittle: this is Webpack build-time variable magic: https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables

src/App.tsx
```diff
@@ -55,6 +56,7 @@ export default class App extends React.Component<IAppProps> {
     }
 
     public render() {
+        console.info(`Running VoTT v${appInfo.version}`);
```
wbreza: Added for prod use or just for local debugging?  Since we already have it in the App Settings page do we need it here?

src/index.scss
```diff
@@ -5,6 +5,10 @@
     overflow: hidden;
 }
 
+#appInfo {
```
wbreza: I'm not seeing this selector used anywhere in the PR.  Still needed?

public/index.html
```diff
@@ -19,13 +19,14 @@
...
     -->
-    <title>React App</title>
+    <title>Visual Object Tagging Tool (VoTT) v%REACT_APP_VERSION%</title>
   </head>
   <body>
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
     <div id="root"></div>
+    <div id="appInfo">VoTT v%REACT_APP_VERSION%, commit=%REACT_APP_COMMIT_SHA%</div>
```
wbreza: How do these %variables% work since they are outside of the React render pipeline? Webpack magic?

src/App.tsx
pjlittle: it's definitely not necessary; was going to fold in more info here (commit, build date, etc.) but turned out to be challenging w/ Webpack. I'm happy to yank. I do like seeing what version I'm using w/ dev tools, but since we're not bumping patch (or something else...) on each PR, it's less useful.

public/index.html
pjlittle: yep, when the app gets built, it does a static replace of these.  I put the link elsewhere, but the 'rules' are that your env var has to start w/ `REACT_APP_` :crystal_ball:

pjlittle:
Removed console log statement. :shipit:

PR #658, opened 2019-03-08
ci: Azure DevOps pipeline definitions for VoTT Web CD

pjlittle:
Yaml files to deploy VoTT to Azure when PRs are merged into
`dev` and `v2` branches.

[AB#17489](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17489)

### DIFF
--- a/azure-pipelines/web/deploy-to-azure-dev.yml
+++ b/azure-pipelines/web/deploy-to-azure-dev.yml
@@ -1,0 +1,18 @@
+trigger:
+- dev
+
+pool:
+  vmImage: 'Ubuntu-16.04'
+
+steps:
+- task: NodeTool@0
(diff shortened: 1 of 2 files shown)


pjlittle:
If there's a way to make this more DRY, I'm all ears... From what I can tell, we need two pipelines setup, to trigger on activities in two different branches.

mydiemho:
> If there's a way to make this more DRY, I'm all ears... From what I can tell, we need two pipelines setup, to trigger on activities in two different branches.

the most you can do is extract the steps to a different file, you'll end up with 3 files instead of 2 but it's "safer" so that if you change the steps, you won't have to worry about forgetting changing the other build.

PR #664, opened 2019-03-10
feat: Added Custom TitleBar & Status Bar

wbreza:
Introduced custom electron frame, titlebar and status bar to enhance overall look and feel of application.

image
image


wbreza:
Confirmed UI is correct with @elizabethhalper after re-pulling branch.

PR #665, opened 2019-03-11
feat: New Tag Editor Component [AB#17094]

tbarlow12:
Here are several screenshots of new functionality

# Selected/locked tag
selected tag
# Editing tag name inline
Edit tag name inline
# Editing tag color inline
Edit tag color inline

Editing is accessible by alt-clicking or selecting a tag and then the edit button on the toolbar.

Re-ordering does not currently support drag and drop, but the arrow buttons should accomplish that.

Resolves bugs [[AB#17358](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17358)] and [[AB#17352](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17352)]


src/common/utils.test.ts
```diff
@@ -51,4 +51,11 @@ describe("Helper functions", () => {
             expect(decryptedProject).toEqual(testProject);
         });
     });
+
+    describe("String utils", () => {
+        it("Elipsifies a string", () => {
+            expect(elipsify("This is a string", 4)).toEqual("This...");
```
wbreza: CSS has native support for elipsing text ex) `text-overflow: ellipsis;`

src/react/components/common/tagInput/tagInputItem.tsx
```diff
@@ -0,0 +1,150 @@
...
+}
+
+function getColorPicker(item, onChange) {
+    return (
+        <div className="tag-color-picker">
+            <GithubPicker
+                color={item.tag.color}
+                onChangeComplete={(color) => handleColorEdit(item.tag, color, onChange)}
+                colors={tagColors}
+                width={165}
```
pjlittle: should these styles/magic number live in our CSS or as consts somewhere else?

tbarlow12: We do have a constants file, that would probably be a good place for them

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -16,6 +16,15 @@
         }
     }
 
+    &-canvas {
+        flex: 1;
+    }
+
+    &-right-sidebar {
+        min-width: 175px;
```
wbreza: Let's use `vw` units if possible

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -122,7 +127,7 @@
             font-size: 80%;
 
             .asset-filename {
-                max-width: 175px;
+                max-width: 200px;
```
wbreza: Why this change?  Seems to cause some overflow in the left sidebar.
image

src/react/components/common/cloudFilePicker/cloudFilePicker.tsx
```diff
@@ -145,6 +145,7 @@ export class CloudFilePicker extends React.Component<ICloudFilePickerProps, IClo
             items={items}
             Component={ListItem}
             onClick={onClick}
+            displayEmptyMessage={true}
```
wbreza: Let's set default values on this prop so we don't have to explicitly go back and make changes in all areas.

src/react/components/common/condensedList/condensedList.tsx
```diff
@@ -15,8 +15,12 @@ interface ICondensedListProps {
     title: string;
     Component: any;
     items: any[];
+    displayEmptyMessage: boolean;
+    Toolbar?: any;
```
wbreza: Why split out the toolbar and the props?  If you want you can just accept a `JSX.element` here and would allow someone to specify their own rendering for this element and pass whatever props they want inline.

wbreza:
**Other UI comments:**
- Condensed list toolbar items need a hover state.
- Color blocks are 2 wide based on Tyler's comps
- Editing a tag and updating text with existing tag name causes funny multiple edit issues
- Color picker should float out to left based on design comps
- Color picker should be a singleton within the Tag input vs. each row having its own
- Each row should have a hover state
- Row selected state gets weird when you assign to a region, most of the time its inverted
- Tag items missing left margin
- Add double click on color box to edit color
- Locked tag icon should float to right side of item

src/react/components/pages/editorPage/editorPage.scss
tbarlow12: This was accidental, must have meant to edit another dimension. Good catch

src/common/utils.ts
```diff
@@ -125,3 +125,27 @@ function decryptProviderOptions<T = IProviderOptions>(providerOptions: IProvider
 
     return decryptObject(providerOptions.encrypted, secret) as T;
 }
+
+export function idealTextColor(background: string) {
```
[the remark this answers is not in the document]
tbarlow12: Agreed, as long as it's the darker tint, white should be fine

tbarlow12:
Closing in favor of new PR

PR #666, opened 2019-03-11
ci: Merge v2 into dev

wbreza:
Merges missing changes from v2 to dev


pjlittle:
I'm assuming these are changes from the transition from `v2` to `dev` as the main working branch? Just some stuff lingering from the cut over?

wbreza:
@pjlittle - I'm assuming we'll want to squash and merge this into 1 commit so we don't end up having multiple duplicate commits

PR #667, opened 2019-03-11
fix: standardize default export option to "visited"

mydiemho:
AB#17532

### DIFF
--- a/src/providers/export/azureCustomVision.json
+++ b/src/providers/export/azureCustomVision.json
@@ -15,7 +15,7 @@
                 "visited",
                 "tagged"
             ],
-            "default": "all",
+            "default": "visited",
             "enumNames": [
                 "${strings.export.providers.common.properties.assetState.options.all}",
                 "${strings.export.providers.common.properties.assetState.options.visited}",
--- a/src/react/components/pages/export/exportForm.tsx
+++ b/src/react/components/pages/export/exportForm.tsx
@@ -160,7 +160,7 @@ export default class ExportForm extends React.Component<IExportFormProps, IExpor
 
         const formData = { ...exportFormat };
         if (resetProviderOptions) {
-            formData.providerOptions = { assetState: ExportAssetState.All };
+            formData.providerOptions = { assetState: ExportAssetState.Visited };
         }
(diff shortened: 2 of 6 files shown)


src/providers/export/azureCustomVision.json
```diff
@@ -15,7 +15,7 @@
                 "visited",
                 "tagged"
             ],
-            "default": "all",
+            "default": "visited",
```
mydiemho: this doesn't seem to make any difference

wbreza: @mydiemho What do you mean?  This should make all the difference regarding how the form controls get initialized.

src/react/components/pages/export/exportForm.tsx
```diff
@@ -160,8 +159,9 @@ export default class ExportForm extends React.Component<IExportFormProps, IExpor
 
         const formData = { ...exportFormat };
         if (resetProviderOptions) {
-            formData.providerOptions = { assetState: ExportAssetState.All };
+            formData.providerOptions = { assetState: ExportAssetState.Visited };
```
mydiemho: > @mydiemho What do you mean? This should make all the difference regarding how the form controls get initialized.

@wbreza  what I mean is that this is the code that set the `Provider Options` to the right default when I switch `Provider Type`.  The `default` in json file doesn't do anything

src/react/components/pages/export/exportForm.tsx
```diff
@@ -160,8 +159,9 @@ export default class ExportForm extends React.Component<IExportFormProps, IExpor
 
         const formData = { ...exportFormat };
         if (resetProviderOptions) {
-            formData.providerOptions = { assetState: ExportAssetState.All };
+            formData.providerOptions = {} as IExportProviderOptions;
```
mydiemho: clear out previous settings and let the form component pull default from the export provider json file

wbreza: Does this work as expected to respect the default values?

mydiemho: > Does this work as expected to respect the default values?

yup it does :D

PR #668, opened 2019-03-11
fix: save asset metadata for all valid asset type [AB#17538]

JacopoMangiavacchi:
- save asset metadata for all valid asset type

Fix [AB#17538]


JacopoMangiavacchi:
fix: save asset metadata for all valid asset type [AB#17538]

PR #669, opened 2019-03-11
fix: display spinner icon on metrics panel when loading

mydiemho:
AB#17500

Screen Shot 2019-03-12 at 11 18 16

### DIFF
--- a/src/react/components/pages/projectSettings/projectMetrics.tsx
+++ b/src/react/components/pages/projectSettings/projectMetrics.tsx
@@ -41,7 +41,6 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
 
     public render() {
         if (this.state.loading) {
-            console.log("Waiting for data");
             return (
                 <div className="project-settings-page-metrics p-3 bg-lighter-1">
                     <h3>
@@ -51,7 +50,7 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
                         </span>
                     </h3>
                     <div className="loading">
-                        <i className="fas fa-spinner fa-6x"/>
+                        <i className="fas fa-circle-notch fa-spin fa-6x"/>
                     </div>
                 </div>
             );
--- a/src/react/components/pages/projectSettings/projectMetrics.tsx
+++ b/src/react/components/pages/projectSettings/projectMetrics.tsx
@@ -42,7 +42,19 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
     public render() {
         if (this.state.loading) {
             console.log("Waiting for data");
-            return <h2> Loading...</h2>;
+            return (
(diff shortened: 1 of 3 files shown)


src/react/components/pages/projectSettings/projectSettingsPage.scss
```diff
@@ -4,11 +4,17 @@
...
 
     &-settings {
-        flex-grow: 1;
+        flex-grow: 2;
     }
 
     &-metrics {
-        align-content: center;
-        align-items: center;
+        flex: 2 1 400px;
```
wbreza: Let's not use physical pixels - instead let's use `vw` units.  See app settings, connection, recent projects SCSS for example.

src/react/components/pages/projectSettings/projectMetrics.tsx
```diff
@@ -42,7 +42,19 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
...
+            return (
+                <div className="project-settings-page-metrics p-3 bg-lighter-1">
+                    <h3>
+                        <i className="fas fa-chart-bar fa-1x"/>
+                        <span className="px-2">
+                            {strings.projectMetrics.title}
+                        </span>
+                    </h3>
+                    <div className="loading">
+                        <i className="fas fa-spinner fa-6x"/>
```
wbreza: Let's standardize on a single spinner icon.  We are using spinner for the `editorSidebar` component `.fa-circle-notch`

src/react/components/pages/projectSettings/projectSettingsPage.scss
```diff
@@ -4,11 +4,17 @@
     height: 100%;
 
     &-settings {
-        flex-grow: 1;
+        flex-grow: 2;
```
wbreza: Why setting this to 2? 1 should be correct. since it is the only element growing within the flexbox.

src/react/components/pages/projectSettings/projectSettingsPage.scss
wbreza: Also consider specifying a min/max width.

src/react/components/pages/projectSettings/projectMetrics.tsx
```diff
@@ -42,7 +42,19 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
     public render() {
         if (this.state.loading) {
             console.log("Waiting for data");
```
pjlittle: do we still need this logging statement?

mydiemho: no I can remove it

src/react/components/pages/projectSettings/projectSettingsPage.scss
mydiemho: i was trying to give the loading metrics panel more space, but I think min-width is a better approach

PR #671, opened 2019-03-11
fix: open file containing pound character in the name [AB#17533]

JacopoMangiavacchi:
- fix: open file containing pound character in the name 

Fix [AB#17533]

### DIFF
--- a/src/react/components/pages/editorPage/editorPage.tsx
+++ b/src/react/components/pages/editorPage/editorPage.tsx
@@ -240,7 +240,7 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
      * Returns a value indicating whether the current asset is taggable
      */
     private isTaggableAssetType = (asset: IAsset): boolean => {
-        return asset.type === AssetType.Image || asset.type === AssetType.VideoFrame;
+        return asset.type !== AssetType.Unknown && asset.type !== AssetType.Video;
     }
 
     /**
--- a/src/services/assetService.ts
+++ b/src/services/assetService.ts
@@ -120,8 +120,9 @@ export class AssetService {
      */
     public async getAssets(): Promise<IAsset[]> {
         // encodeURI() will not encode: ~!@#$&*()=:/,;?+'
-        // extend it to support all of these except : and /
-        const natchString = "/#/g";
+        // extend it to support all of these except # and ?
+        // all other non encoded characters are implicitly supported with no reason to encoding them
+        const matchString = /#/g;
         const encodings = {
             "\#": "%23",
           };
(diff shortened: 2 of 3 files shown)


src/services/assetService.ts
```diff
@@ -129,8 +129,9 @@ export class AssetService {
             if (!normalizedPath.startsWith("http://") &&
                 !normalizedPath.startsWith("https://") &&
                 !normalizedPath.startsWith("file:")) {
-                asset.path = "file:" + asset.path;
-                asset.path = encodeURI(asset.path.replace(/\\/g, "/"));
+                let path = encodeURI("file:" + asset.path.replace(/\\/g, "/"));
+                path = path.replace(/#/g, "%23");
+                asset.path = path;
```
wbreza: Instead of 1 off special case solution I would rather we refactor this to use `encodeURIComponent(...)` method to encode the filename portion of the path.  The fix may need to be inside of `AssetService.createAssetFromFilePath`

https://stackoverflow.com/questions/4540753/should-i-use-encodeuri-or-encodeuricomponent-for-encoding-urls

I would also like to see some unit tests for these bug fixes.

JacopoMangiavacchi: Actually I see encodeURIComponent will encode everything, slashes included.  It seams to be an api recomended for encoding query strings and not entire URI.
`var uri = 'http://example.com/foo?hello=' + encodeURIComponent(world);`

The problem with encodeURI, that instead it assumes that the input is a complete URI, is that it doesn't encode ~!@#$&*()=:/,;?+'

I found this escamotage for encodeURI here: https://stackoverflow.com/questions/6627648/how-can-i-submit-a-hash-key-in-an-url-parameter

Anyway I think best method to call in our case should be **escape()**

JacopoMangiavacchi: escape is encoding unfortunately the : character you may have on windows path

JacopoMangiavacchi: I'm implementing a custom encoding table and generilized regular expression to handle all exception cases from encodeURI() that we want to handle.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -240,7 +240,7 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
      * Returns a value indicating whether the current asset is taggable
      */
     private isTaggableAssetType = (asset: IAsset): boolean => {
-        return asset.type === AssetType.Image || asset.type === AssetType.VideoFrame;
+        return asset.type !== AssetType.Unknown && asset.type !== AssetType.Video;
```
elizabethhalper: Is an asset with AssetType === Video not considered taggable?

JacopoMangiavacchi: That's right. Video is the parent but the real taggable assets are his child of type VideoFrame

PR #673, opened 2019-03-12
fix: Re-apply tags in editor when project tags are updated

wbreza:
Resolves issue where if tag color is modified while in the editor page, the color changes are not applied till you reload or navigate away from the current asset

Resolves AB#17264

### DIFF
--- a/src/react/components/pages/editorPage/canvas.test.tsx
+++ b/src/react/components/pages/editorPage/canvas.test.tsx
@@ -16,9 +16,9 @@ import { RegionsManager } from "vott-ct/lib/js/CanvasTools/Region/RegionsManager
 import Confirm, { IConfirmProps } from "../../common/confirm/confirm";
 import { Rect } from "vott-ct/lib/js/CanvasTools/Core/Rect";
 import { SelectionMode } from "vott-ct/lib/js/CanvasTools/Interface/ISelectorSettings";
+import { wrap } from "module";
 
 describe("Editor Canvas", () => {
-
     function createComponent(canvasProps?: ICanvasProps, assetPreviewProps?: IAssetPreviewProps)
         : ReactWrapper<ICanvasProps, ICanvasState, Canvas> {
         const props = createProps();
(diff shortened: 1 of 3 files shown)


src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -568,11 +568,26 @@ describe("Editor Canvas", () => {
...
+        const project = { ...wrapper.props().project };
+        const tags = [...project.tags];
+        tags[0].color = "#FFCCFF";
+        project.tags = tags;
+
+        const editor = wrapper.instance().editor;
+        const updateTagsSpy = jest.spyOn(editor.RM, "updateTagsById");
+
+        wrapper.setProps({ project });
+        expect(updateTagsSpy).toBeCalledTimes(assetMetadata.regions.length);
```
mydiemho: I feel like this is exposing/dependent on internal implementation.  How about testing that all the regions have a new tag color?

src/react/components/pages/editorPage/canvas.test.tsx
```diff
@@ -512,7 +512,7 @@ describe("Editor Canvas", () => {
                 ...cProps.selectedAsset,
                 regions: [copiedRegion],
             },
-        }).find(Canvas);
```
mydiemho: how does the tests still pass with this gone / how does this not break tests before?

wbreza: The `wrapper` is the canvas and this isn't needed.  If you take a look at `createComponent()` the root element that it returns is the `Canvas` component itself.

src/react/components/pages/editorPage/canvas.test.tsx
wbreza: Our tag color is modified directly by the tags input.  This bug that the canvas component didn't update to reflect tag updates.  Extending this test beyond calling into canvas tools API would be re-testing the internal implementation of the canvas itself.

PR #678, opened 2019-03-13
feat: Updated UX of project metrics to include charts

wbreza:
Fix includes adding `react-vis` to display charts for assets & tags metrics.

image

Resolve AB#17535


src/react/components/pages/projectSettings/projectMetrics.tsx
```diff
@@ -145,12 +144,8 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
         const sourceAssets = await assetService.getAssets();
 
         const assetsMap = this.props.project.assets;
```
mydiemho: could `project.assets` be null?

wbreza: It shouldn't be. It is always initialized to a empty object.

PR #680, opened 2019-03-13
refactor: remove redux-invariant & redux-logging from production

mydiemho:
these middlewares are intend to be use only during development.

[AB#16163](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=16163)


src/common/environment.ts
```diff
@@ -2,4 +2,8 @@ export class Env {
     public static get() {
         return process.env.NODE_ENV;
     }
+
+    public static isDevelopment(): boolean {
```
tbarlow12: This may be a little misleading. There are more envs than just dev and prod (I know there's test, but there may be more?). Maybe instead something like `isProd` and then `!` it when running the check

package.json
```diff
@@ -108,6 +106,8 @@
         "popper.js": "^1.14.6",
         "react-jsonschema-form": "^1.0.6",
```
wbreza: while you are in here - these 2 should be real dependencies
- react-jsonschema-form
- react-modal

src/common/environment.ts
wbreza: Can we add a test for this?  Also - did you confirm that will work as expected in web target?

wbreza: I would also only add these middlewares in dev.  So that in tests and prod they aren't registered.  In test run they log a lot of unneeded noise.

src/redux/store/store.ts
```diff
@@ -1,27 +1,37 @@
...
  */
 export default function createReduxStore(
     initialState?: IApplicationState,
     useLocalStorage: boolean = false): Store {
     const paths: string[] = ["appSettings", "connections", "recentProjects"];
-    const middlewares = useLocalStorage
-        ? applyMiddleware(thunk, reduxImmutableStateInvarient(), createLogger(), createLocalStorage({ paths }))
-        : applyMiddleware(thunk, reduxImmutableStateInvarient(), createLogger());
+
+    const middlewares = [];
```
wbreza: You'll need to write a unit test or 2 to ensure we meet the code coverage requirements.

PR #682, opened 2019-03-13
fix: replace default react favicon with tags icon

elizabethhalper:
Added new favicon.ico file to public folder in order to have "VoTT" icon display in tabs. Tested in Chrome, Firefox, and Safari.


wbreza:
Need to also update the icons in the `app-icons` folder.

PR #690, opened 2019-03-19
fix: Don't call tag click with invalid hot key [AB#17639]

tbarlow12:


src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -518,7 +518,31 @@ describe("Editor Page Component", () => {
 
             expect(editorPage.state().lockedTags).toEqual([]);
             expect(editorPage.state().selectedTag).toEqual(project.tags[0].name);
+        });
+
+        it("Does not set selected tag or locked tags when invalid hot key is pressed", async () => {
+            const project = MockFactory.createTestProject();
+            const store = createReduxStore({
+                ...MockFactory.initialState(),
+                currentProject: project,
```
mydiemho: how many tags are in the project? we should make that clear in the mock

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -518,7 +518,31 @@ describe("Editor Page Component", () => {
 
             expect(editorPage.state().lockedTags).toEqual([]);
             expect(editorPage.state().selectedTag).toEqual(project.tags[0].name);
+        });
+
+        it("Does not set selected tag or locked tags when invalid hot key is pressed", async () => {
```
mydiemho: we should have 2 tests here:

1) selecting new tag will unlock all existing locked tags
2) using a non-existent hot-key won't do anything (locked tags are still locked, no new tag selected)

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -510,15 +510,29 @@ describe("Editor Page Component", () => {
...
-            const editorPage = wrapper.find(EditorPage).childAt(0);
+            expect(editorPage.state().selectedTag).toBeNull();
 
             dispatchKeyEvent("1");
 
-            expect(editorPage.state().lockedTags).toEqual([]);
             expect(editorPage.state().selectedTag).toEqual(project.tags[0].name);
+        });
+
+        it("does not set selected tag or locked tags when hot key is pressed", async () => {
```
mydiemho: clarified as `non-valid` hot key

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -501,7 +501,7 @@ describe("Editor Page Component", () => {
         });
 
         it("sets selected tag and locked tags when hot key is pressed", async () => {
```
mydiemho: since we're not testing anything relating to locked tags here, we should update the test description

src/react/components/pages/editorPage/editorPage.test.tsx
mydiemho: test not related to log tags, we should either update test or update description

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -611,6 +625,33 @@ describe("Editor Page Component", () => {
             editorPage = wrapper.find(EditorPage).childAt(0);
             expect(editorPage.state().lockedTags).toEqual([]);
         });
+
+        it("Clears locked tags when tag is clicked", async () => {
```
mydiemho: add an identical one but for when hot-key selection is use

tbarlow12: So the way we are simulating the events (click simulations vs dispatching of key events) makes it pretty difficult to combine the two simulations. In order to set locked tags to begin with, I need to simulate a click event that sets the locked tags. I can't set them directly in state, because the editor page is not the root component. But I think it's adequate to show the selected tag is changing because those are being set by the same call to `setState`

PR #701, opened 2019-03-21
fix: Catch error while tfrecord image not loaded

PIC123:
Fixes bug where tfrecords show error loading image

Resolves [AB#17680](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17680)

### DIFF
--- a/src/react/components/common/assetPreview/tfrecordAsset.tsx
+++ b/src/react/components/common/assetPreview/tfrecordAsset.tsx
@@ -28,7 +28,7 @@ export class TFRecordAsset extends React.Component<IAssetProps, ITFRecordState> 
             <img ref={this.image}
                 src={this.state.tfRecordImage64}
                 onLoad={this.onLoad}
-                onError={this.props.onError} />
+                onError={this.onError} />
         );
     }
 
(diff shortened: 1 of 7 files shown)


src/react/components/common/assetPreview/tfrecordAsset.tsx
```diff
@@ -54,6 +54,12 @@ export class TFRecordAsset extends React.Component<IAssetProps, ITFRecordState>
         }
     }
 
+    private onError = (e: React.SyntheticEvent<Element>) => {
+        if (this.props.onError && this.state.tfRecordImage64 !== "") {
```
tbarlow12: Is there a way to check that `e` itself is empty? I thought you mentioned that the error had no message

PIC123: So I actually don't think that was the issue, but rather that we should ignore the error initially when the asset preview tries to load an empty string.  The error is correct still I think and we should catch if the tfrecord image can't load, just not because of an empty string.

PR #707, opened 2019-03-25
ci: refactor plato report pipeline

mydiemho:
1. replace `az storage` command with `az copy` task for uploading
1. templatize pipeline

[AB#17780](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17780)

Generate report (build) => Upload (release)
https://dev.azure.com/msft-vott/VoTT/_build/results?buildId=1961 => https://dev.azure.com/msft-vott/VoTT/_releaseProgress?_a=release-environment-logs&releaseId=167&environmentId=167

### DIFF
--- a/azure-pipelines/plato/generate-up-to-date-report.yml
+++ b/azure-pipelines/plato/generate-up-to-date-report.yml
@@ -42,7 +42,7 @@ steps:
     versionSpec: 10.x
 
 - task: Npm@1
-  displayName: 'npm custom'
+  displayName: 'Run `npm ci`'
   inputs:
     command: custom
     verbose: false
--- a/azure-pipelines/plato/generate-up-to-date-report.yml
+++ b/azure-pipelines/plato/generate-up-to-date-report.yml
@@ -1,0 +1,69 @@
+name: $(Date:yyyyMMdd).$(Hours)$(Minutes)$(Seconds)
+variables:
+  AZURE_STORAGE_ACCOUNT: vottv2
+trigger:
+  - dev*
+  - v2
+pr: none # disable CI build for PR
+
+pool:
+  vmImage: 'Ubuntu-16.04'
+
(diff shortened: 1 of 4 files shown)


scripts/update-report.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -euo pipefail
+set -eo pipefail
```
pjlittle: any reason to yank `u` here?

azure-pipelines/plato/generate-up-to-date-report.yml
```diff
@@ -0,0 +1,69 @@
...
+    inlineScript: './scripts/update-report.sh'
+
+- bash: |
+    set -e
+
+    cat /tmp/download.log
+  displayName: "print download log"
+  condition: succeededOrFailed()
+
+- task: PublishBuildArtifacts@1
```
pjlittle: is this the new copy task that pushes to blob?

mydiemho: no, this is just to publish the report as an artifact, which is then use in this release to copy to blob

https://dev.azure.com/msft-vott/VoTT/_releaseProgress?_a=release-environment-logs&releaseId=167&environmentId=167

scripts/update-report.sh
mydiemho: this was removed when I was adding some code to access pipeline variables in one of the iteration, I can put it back.

Script will fail if we ever try to access pipelines variables because it will always be unset
```
variable=$(SOME_VARIABLE)
```

pjlittle: got it - usually desirable behavior, but I can see how it could be a hinderance in this case.

PR #709, opened 2019-03-26
feat: Update asset status to use badges & resize sidebar

wbreza:
Updates the green/yellow shading on assets to use badges to display on top of asset.  Also adds resizing to sidebar to allow the user to select the size of the asset sidebar which is persisted into appSettings.

**Small**
image

**Medium**
image

**Large**
image


src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -251,5 +233,76 @@ canvas {
 }
 
 .asset-error {
+    font-size: 85%;
+    font-weight: 500;
     margin: auto;
 }
+
+.badge-visited, .badge-tagged {
```
mydiemho: is it a CSS thing to not just call this `.badge` and have the `tagged` and `visisted` inherited from it?

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -35,7 +37,9 @@ import { strings } from "../../../../common/strings";
 export interface IEditorPageProps extends RouteComponentProps, React.Props<EditorPage> {
     project: IProject;
     recentProjects: IProject[];
+    appSettings: IAppSettings;
```
mydiemho: missing jsdoc

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -60,18 +64,21 @@ export interface IEditorPageState {
     selectedTag: string;
     /** Tags locked for region labeling */
     lockedTags: string[];
+    thumbnailSize: ISize;
```
mydiemho: missing comment

Nit: we should refactor existing comment into proper jsdoc

src/react/components/pages/editorPage/editorSideBar.tsx
```diff
@@ -13,6 +13,7 @@ export interface IEditorSideBarProps {
     assets: IAsset[];
     onAssetSelected: (asset: IAsset) => void;
     selectedAsset?: IAsset;
+    thumbnailSize?: ISize;
```
mydiemho: update jsdoc

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -251,5 +233,76 @@ canvas {
...
+    color: #fff;
+    position: absolute;
+    z-index: 2;
+    right: 0.5em;
+    top: 0.5em;
+    text-shadow: 0px 0px 4px #000
```
luisamiranda: 1px or 2px shadow is more in keeping with the rest of the icons - just enough to hold the edges.

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -251,5 +233,76 @@ canvas {
...
+}
+
+.badge-tagged {
+    background-color: rgba(green, 0.9);
+    border: 1px solid $lighter-2;
+}
```
luisamiranda: remove border for uniformity with other icons

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -251,5 +233,76 @@ canvas {
...
+
+.badge-visited {
+    background-color: rgba(darkorange, 0.9);
+    border: 1px solid $lighter-3;
```
luisamiranda: remove border for uniformity with other icons

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -251,5 +233,76 @@ canvas {
...
+    color: #fff;
+    position: absolute;
+    z-index: 2;
+    right: 0.5em;
+    top: 0.5em;
+    text-shadow: 0px 0px 4px #000
+}
+
+.badge-tagged {
+    background-color: rgba(green, 0.9);
```
luisamiranda: consider using same green as logo - #7cbb00

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -251,5 +233,76 @@ canvas {
...
+    font-size: 85%;
+    font-weight: 500;
     margin: auto;
 }
+
+.badge-visited, .badge-tagged {
+    color: #fff;
+    position: absolute;
+    z-index: 2;
+    right: 0.5em;
```
luisamiranda: right: 0; looks better to me

src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -251,5 +233,76 @@ canvas {
...
+    top: 0.5em;
```
luisamiranda: top: 0; (or just remove line) looks better on smaller size, imo

src/react/components/pages/editorPage/editorPage.scss
wbreza: There is already a `badge` class in bootstrap that we are leveraging and bot of these classes have the same settings.

src/react/components/pages/editorPage/editorPage.scss
wbreza: `right: 0` is right up against the edge and doesn't look right to me.  Keeping with `0.5em` for now.

src/react/components/pages/editorPage/editorPage.scss
wbreza: Same comment as above - needs some space from corner.

src/react/components/pages/editorPage/editorPage.scss
wbreza: This is more of a glow radius vs an actual shadow to allow the white icon to stand out better on top of the lighter (orange) badge color.

src/react/components/pages/editorPage/editorPage.scss
wbreza: I initially tried that but like how the darker green looks.

src/react/components/pages/editorPage/editorPage.scss
wbreza: most of other icons are using in a toolbar and I would agree with that.  This border is on the badge, not the icon and feel like it needs a little separation from the image since its floating on top.  Its a very faint border only slightly lighter than the badge color itself.

src/react/components/pages/editorPage/editorPage.scss
wbreza: Same as above.

src/react/components/pages/editorPage/editorPage.tsx
wbreza: These are proper JS doc comments for the properties and the Storybook JS doc plugin expect it in this format.  Added missing doc.

PR #711, opened 2019-03-27
feat: v1 video import

elizabethhalper:
Add ability to import v1 projects made with a video asset


src/react/components/pages/homepage/homePage.tsx
```diff
@@ -185,8 +188,11 @@ export default class HomePage extends React.Component<IHomePageProps, IHomePageS
 
         this.props.applicationActions.ensureSecurityToken(project);
 
+        if (project.lastVisitedAssetId !== undefined) {
```
wbreza: Can be simplified to `if(project.lastVisitedAssetId)` since any `null` or `undefined` value will evaluate to false.

src/react/components/pages/homepage/homePage.tsx
```diff
@@ -197,7 +203,21 @@ export default class HomePage extends React.Component<IHomePageProps, IHomePageS
...
+
+        // const rootAsset = { ...(assetMetadata.asset.parent || assetMetadata.asset) };
+
         await this.props.actions.saveProject(this.props.project);
+
+        if (parent) {
+            const assetService = new AssetService(this.props.project);
+            const childAssets = assetService.getChildAssets(parent);
+
+            console.log(childAssets);
```
wbreza: Remove console.log

src/react/components/pages/homepage/homePage.tsx
wbreza: I don't think anything specific to a given asset image/video should leak up into this part of the implementation.  This feels like something that should be handled by the `ImportService` that you are calling below.

src/services/importService.ts
```diff
@@ -61,33 +65,61 @@ export default class ImportService implements IImportService {
             },
             autoSave: true,
         };
+
+        if (originalProject.visitedFrames.every((item) => typeof item === "number")) {
+            // then this is a video project
+            convertedProject.lastVisitedAssetId = "dummyId";
```
wbreza: Why setting `lastVisitedAssetId` to this value?  This is an optional property and if we don't have a real value to set it to we shouldn't bother setting it.

src/services/importService.ts
```diff
@@ -96,6 +128,30 @@ export default class ImportService implements IImportService {
...
+     */
+    public async createParentVideoAsset(v1Project: IFileInfo): Promise<IAsset> {
+        let parentAsset: IAsset;
+        // parentAsset = AssetService.createAssetFromFilePath(v1Project.file.path);
+        parentAsset = {
+            format: "mp4",
+            id: new MD5().update(v1Project.file.path.replace(/\.[^/.]+$/, "")).digest("hex"),
+            name: v1Project.file.path.replace(/\.[^/.]+$/, "").replace(/^.*[\\\/]/, ""),
+            path: `file:${v1Project.file.path.replace(/\.[^/.]+$/, "")}`,
+            size: {
```
wbreza: We should be pulling actual width/height values if available in the source project.

src/services/importService.ts
```diff
@@ -61,33 +65,61 @@ export default class ImportService implements IImportService {
...
 
         for (const frameName in originalProject.frames) {
             if (originalProject.frames.hasOwnProperty(frameName)) {
                 const frameRegions = originalProject.frames[frameName];
-                const asset = AssetService.createAssetFromFilePath(
-                    `${v1Project.file.path.replace(/[^\/]*$/, "")}${frameName}`);
-                const populatedMetadata = await assetService.getAssetMetadata(asset).then((metadata) => {
+                let asset: IAsset;
+                // if video:
+                if (originalProject.visitedFrames.every((item) => typeof item === "number") && parent !== null) {
```
wbreza: I recommend splitting this function up into multiple smaller functions to handle the different between images & video.

src/services/importService.ts
```diff
@@ -61,33 +65,61 @@ export default class ImportService implements IImportService {
...
+
         return convertedProject;
     }
 
     /**
      * Generate assets based on V1 Project frames and regions
      * @param project - V1 Project Content and File Information
      */
-    public async generateAssets(v1Project: IFileInfo, v2Project: IProject): Promise<IAssetMetadata[]> {
+    public async generateAssets(v1Project: IFileInfo, v2Project: IProject, parent?: IAsset): Promise<IAssetMetadata[]> {
```
wbreza: I suggest we don't pull in the top level video asset here.  The code path for image project or video project should be treated the same in my opinion and the inspection of the project file should happen within this function implementation.

src/services/importService.ts
```diff
@@ -70,32 +74,48 @@ export default class ImportService implements IImportService {
...
 
+    /**
+     * Generate parent asset based on V1 Project video assets
+     * @param project - V1 Project Content and File Information
+     */
+    private async createParentVideoAsset(v1Project: IFileInfo): Promise<IAsset> {
+        let parentAsset: IAsset;
+
+        parentAsset = {
+            format: "mp4",
```
wbreza: We cannot assume this will always be a mp4 file.

PR #713, opened 2019-03-29
feat: exclude desktop functionality from browser target

pjlittle:
Remove electron functionality from browser builds. Some updates to
CI scripts, based on things found in the last release.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -euo pipefail
+
+# NOTE: this script should be ran from the root of the repository; the CWD should reflect this
+VERSION=$(node -pe "require('./package.json').version")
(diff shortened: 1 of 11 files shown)


scripts/build.sh
```diff
@@ -0,0 +1,18 @@
...
+NPM_BIN_DIR=$(npm bin)
+
+echo "cwd=$(pwd)"
+echo "version=${VERSION}"
+echo "commit=${COMMIT_SHA}"
+
+export REACT_APP_VERSION=${VERSION}
+export REACT_APP_COMMIT_SHA=${COMMIT_SHA}
+
+npx react-scripts build
```
pjlittle: was reminded that npx is a super handy way to run bin/cli tools w/o having to reference into `node_modules`.

PR #715, opened 2019-04-01
fix: CORS warning message for Bing Image Search

pjlittle:
Error message displays when adding a Bing Image Search asset provider
type. This is only applicable for the Browser target.

[AB#17153](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17153)

### DIFF
--- a/src/common/localization/en-us.ts
+++ b/src/common/localization/en-us.ts
@@ -163,6 +163,12 @@ export const english: IAppStrings = {
             saveSuccess: "Successfully saved ${connection.name}",
             deleteSuccess: "Successfully deleted ${connection.name}",
         },
+        imageCorsWarning: "Warning: When using VoTT in a Web browser, some assets from Bing Image \
+                          Search may not export correctly due to CORS (cross origin resource sharing) restrictions.",
+        blobCorsWarning: "Warning: When using VoTT in a Web browser, assets in Azure Blob Storage will not export  \
+                         if CORS is not enabled. For more information on enabling CORS can be found \
+                         in the {0}",
+        azDocLinkText: "Azure Documentation.",
         providers: {
             azureBlob: {
                 title: "Azure Blob Storage",
--- a/src/common/localization/es-cl.ts
+++ b/src/common/localization/es-cl.ts
@@ -164,6 +164,13 @@ export const spanish: IAppStrings = {
             saveSuccess: "${connection.name} guardado correctamente",
             deleteSuccess: "${connection.name} eliminado correctamente",
         },
+        imageCorsWarning: "Advertencia: Cuando se usa VoTT en un navegador web, es posible que algunos activos de este \
+                          Búsqueda de Imágenes Bing no se exporten correctamente debido a las restricciones de CORS \
+                          (recursos de origen cruzado).",
+        blobCorsWarning: "Advertencia: Cuando se usa VoTT en un navegador web, los activos en Azure Blob \
+                         Storage no se exportarán si CORS no está habilitado. Para obtener más información sobre \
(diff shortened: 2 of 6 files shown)


src/common/localization/en-us.ts
```diff
@@ -143,6 +143,8 @@ export const english: IAppStrings = {
         settings: "Connection Settings",
         instructions: "Please select a connection to edit",
         save: "Save Connection",
+        corsWarning: "Warning: when using VoTT in the browser, assets from this provider may not load correctly \
```
pjlittle: If someone has better copy here, I'm all ears!

src/react/components/pages/connections/connectionForm.tsx
```diff
@@ -85,6 +86,12 @@ export default class ConnectionForm extends React.Component<IConnectionFormProps
                     </span>
                 </h3>
                 <div className="m-3 text-light">
+                    { isBrowser() && this.state.providerName === "bingImageSearch" &&
```
pjlittle: is there a better way to compare for Bing Image? providerName (type) is only define in the register script. Is there another type comparison that can be done here? I don't want to use the display name.

src/common/localization/en-us.ts
wbreza: This issue is with exporting vs. loading.  Maybe something to the effect of "Warning: When using VoTT in the browser some assets from this provider may not export correctly due to CORS (cross origin resource sharing) restrictions.

src/react/components/pages/connections/connectionForm.tsx
wbreza: Nothing that I can think of right now - this is the first time we've wanted to add some additional text inline within the provider form.  Only other thing would be to define a "description" in the provider settings, but it would show for all host modes.

src/react/components/pages/connections/connectionForm.tsx
```diff
@@ -85,6 +86,12 @@ export default class ConnectionForm extends React.Component<IConnectionFormProps
...
+                    <div className="alert alert-warning" role="alert">
+                        <i className="fas fa-exclamation-circle" aria-hidden="true"></i>
```
wbreza: You can use bootstrap build in classes like `mr-1` which will add margin-right to the element vs defining a new css class.

pjlittle: ah, nice, will do.

src/common/localization/en-us.ts
pjlittle: got it - I thought it was mainly around being able to display at all. I'll update to this.  One more question: Should we do something similar for blob, or just call it good? I'm leaning towards the latter at this point...

wbreza: We have something in the help docs about enabling CORS in blob, but it would be better if there was a message inline within the app since that would limit actual loading of images.

src/common/localization/es-cl.ts
```diff
@@ -144,6 +144,9 @@ export const spanish: IAppStrings = {
         details: "Detalles de Conexión",
         settings: "Configuración de Conexión",
         instructions: "Por favor seleccione una conexión para editar",
+        corsWarning: "Advertencia: cuando se usa VoTT en un navegador web, es posible que algunos activos de este \
```
tbarlow12: Nit: Capitalize `cuando`

src/common/localization/en-us.ts
pjlittle: Since I went back to fix up the translation, I went ahead and did this for blob as well.

PR #717, opened 2019-04-03
fix: switch asset count to include video frames

PIC123:
Switch project metrics to track video frames as assets when calculating total assets


src/react/components/pages/projectSettings/projectMetrics.tsx
```diff
@@ -210,7 +210,14 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
      *   Note: video frames are not counted, only the video container
      */
     private getSourceAssetCount = () => {
-        const assets = this.state.sourceAssets;
+        const assets = this.state.projectAssetsMetadata.map((e) => e.asset.name);
```
wbreza: Consider using JS `Set<IAsset>` here to help deal with duplicates.  You would avoid having to check for existence before adding assets.

wbreza:
@PIC123 - Couple minor UI things to sort out
- Hover popup wraps and renders weird. Need more spacing after color cube
- Need margin between the 2 charts, they are very close together
- Stilling seeing negative value in one of my projects (Not Visited Assets)

src/react/components/pages/projectSettings/projectMetrics.tsx
```diff
@@ -31,16 +36,32 @@ export interface IProjectMetricsState {
 export default class ProjectMetrics extends React.Component<IProjectMetricsProps, IProjectMetricsState> {
     public state = {
         loading: true,
+        hoveredCell: null,
         sourceAssets: [],
         projectAssetsMetadata: [],
     };
 
+    private tipStyle = {
```
pjlittle: can the inline styles live in a `scss` file or typical for this component to use them like this?

src/react/components/pages/projectSettings/projectMetrics.tsx
```diff
@@ -64,26 +85,68 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
...
-                angle: taggedAssetCount,
-                label: interpolate(strings.projectMetrics.taggedAssets, { count: taggedAssetCount }),
-            },
-            {
-                angle: nonVistedAssetCount,
-                label: interpolate(strings.projectMetrics.nonVisitedAssets, { count: nonVistedAssetCount }),
-            },
-        ];
+        const visitedAssetCount = this.getVisitedAssetsCount();
+        const assetChartSize = window.innerWidth >= 1920 ? 250 : 200;
```
pjlittle: do we use this number (1920) anywhere else? Seems like a good const candidate, but perhaps this is the only usage. Often time 'magic numbers' can be hard to track down later :)

pjlittle:
Lots of styles inline, but perhaps par for the course for using this component.

src/react/components/pages/projectSettings/projectMetrics.tsx
wbreza: Good catch - this is the only place where we are doing something like this right now.  We can definitely move out to a constant though.

src/react/components/pages/projectSettings/projectMetrics.tsx
```diff
@@ -94,22 +157,65 @@ export default class ProjectMetrics extends React.Component<IProjectMetricsProps
...
+                        getLabel={(d) => d.name}
+                        getSize={(d) => d.size}
+                        getColor={(d) => d.clr}
+                        width={assetChartSize}
+                        padAngle={() => 0.05}
+                        hideRootNode={true}
+                    >
+                        {hoveredCell ? (
+                            <Hint value={this.buildValue(hoveredCell)}>
+                                <div style={this.tipStyle}>
```
wbreza: Convert these to CSS class name references.

src/react/components/pages/projectSettings/projectMetrics.tsx
wbreza: Agreed - these should be moved to SCSS file.

PR #719, opened 2019-04-03
fix: Render Enhancments

wbreza:
Resolves several fit and finish issues

- Adds missing localization mesages
- Re-floats tag panel to the far right vs inside the body (see screenshot)
- Add css breakpoints to hide titlebar menu on small screen
- Adds css to provider better UX on small screen
- Adds checkbox widget to Azure blob storage

Resolves [AB#18091](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=18091)

image

### DIFF
--- a/src/electron/main.ts
+++ b/src/electron/main.ts
@@ -16,6 +16,8 @@ function createWindow() {
         height: 768,
         frame: process.platform === "linux",
         titleBarStyle: "hidden",
+        backgroundColor: "#272B30",
+        show: false,
     };
 
     const staticUrl = process.env.ELECTRON_START_URL || `file:///${__dirname}/index.html`;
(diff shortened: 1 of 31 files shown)


src/electron/main.ts
```diff
@@ -16,6 +16,8 @@ function createWindow() {
         height: 768,
         frame: process.platform === "linux",
         titleBarStyle: "hidden",
+        backgroundColor: "#272B30",
```
wbreza: Reduces flash of white screen - set's app background color to our main color.

src/electron/main.ts
```diff
@@ -37,6 +39,12 @@ function createWindow() {
         mainWindow = null;
     });
 
+    // Provides a more graceful experience and eliminates the white screen on load
```
wbreza: Also helps to reduce white screen on app launch - Only show the main window after the first app render

PR #728, opened 2019-04-06
feat: Adds option to specify if images are included with JSON export

wbreza:
Adds option that allows the user to specify whether or not binary images assets will be generated as part of the JSON export provider

Resolves [AB#18068](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=18068)

### DIFF
--- a/src/common/localization/es-cl.ts
+++ b/src/common/localization/es-cl.ts
@@ -280,8 +280,8 @@ export const spanish: IAppStrings = {
                 displayName: "VoTT JSON",
                 properties: {
                     includeImages: {
-                        title: "Include Images",
-                        description: "Whether or not to include binary image assets in target connection",
+                        title: "Incluir imágenes",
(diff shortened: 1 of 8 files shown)


src/common/localization/es-cl.ts
```diff
@@ -278,6 +278,12 @@ export const spanish: IAppStrings = {
             },
             vottJson: {
                 displayName: "VoTT JSON",
+                properties: {
+                    includeImages: {
+                        title: "Include Images",
```
tbarlow12: Incluir Imagenes

src/common/localization/es-cl.ts
```diff
@@ -278,6 +278,12 @@ export const spanish: IAppStrings = {
...
+                        description: "Whether or not to include binary image assets in target connection",
```
tbarlow12: Si se incluyen o no los activos de imágenes binarias en la conexión de destino

PR #732, opened 2019-04-09
fix: CORS warning message for Bing Image Search

pjlittle:
Error message displays when adding a Bing Image Search asset provider
type. This is only applicable for the Browser target.

[AB#17153](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=17153)

### DIFF
--- a/src/common/localization/en-us.ts
+++ b/src/common/localization/en-us.ts
@@ -163,6 +163,12 @@ export const english: IAppStrings = {
             saveSuccess: "Successfully saved ${connection.name}",
             deleteSuccess: "Successfully deleted ${connection.name}",
         },
+        imageCorsWarning: "Warning: When using VoTT in a Web browser, some assets from Bing Image \
+                          Search may not export correctly due to CORS (Cross Origin Resource Sharing) restrictions.",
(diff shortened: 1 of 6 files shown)


pjlittle:
NOTE: same code, different PR -- the build/status checks were hanging from last week.

src/common/localization/en-us.ts
```diff
@@ -163,6 +163,12 @@ export const english: IAppStrings = {
             saveSuccess: "Successfully saved ${connection.name}",
             deleteSuccess: "Successfully deleted ${connection.name}",
         },
+        imageCorsWarning: "Warning: When using VoTT in a Web browser, some assets from Bing Image \
+                          Search may not export correctly due to CORS (cross origin resource sharing) restrictions.",
+        blobCorsWarning: "Warning: When using VoTT in a Web browser, assets in Azure Blob Storage will not export  \
```
mydiemho: is this about using Blob as source or target connection?  the wording is a bit confusing

wbreza: if you don't setup CORS you won't be able to use azure blob storage as a source or a target.

PR #734, opened 2019-04-10
feat: Auto-size video/image assets and position canvas

wbreza:
Normalizes the display of both image and video assets in VoTT.  `<video>` tags by default display with `object-fit: contain` which forces the video to always render in its correct aspect ratio.  The browser will automatically add horizontal / vertical bars as necessary based on fluid layouts.  This updates images to also use this same technique to preserve aspect ratio.  

The side effect however is having to manually calculate the actual asset content from within the HTML elements by detecting aspect ratio and computing correct offsets.

### DIFF
--- a/src/react/components/pages/editorPage/editorPage.scss
+++ b/src/react/components/pages/editorPage/editorPage.scss
@@ -60,7 +60,7 @@
 
       &-video {
         position: static;
-        width: auto;
+        width: 100%;
         height: calc(100% - 3em);
       }
 
@@ -161,9 +161,9 @@
   }
 
   img {
-    margin: auto;
-    max-width: 100%;
-    max-height: 100%;
+    width: 100%;
+    height: 100%;
+    object-fit: contain;
   }
 
   &.portrait {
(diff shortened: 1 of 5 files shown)


src/react/components/pages/editorPage/editorPage.scss
```diff
@@ -226,6 +226,7 @@
 #editor-zone {
   width: auto !important;
   padding: 0 !important;
+  height: 100% !important;
```
wbreza: Resolves an awkward calculation made by canvas tools and allow VoTT to regain control over height of element.

wbreza:
Normalizes the display of both image and video assets in VoTT. <video> tags by default display with object-fit: contain which forces the video to always render in its correct aspect ratio. The browser will automatically add horizontal / vertical bars as necessary based on fluid layouts. This updates images to also use this same technique to preserve aspect ratio.

The side effect however is having to manually calculate the actual asset content from within the HTML elements by detecting aspect ratio and computing correct offsets.

PR #736, opened 2019-04-10
feat: Active Learning

JacopoMangiavacchi:
This PR enable Active Learning functionalities deploying in the app bundle a pre-trained Mobilenet-Coco-SSD object detection model that could be used to simply detect regions and optionally associate tags to these detected regions from the ~90 classes provided from the MS-Coco training dataset.

New Project settings configuration allows to customize AL settings and to download a specifically trained model from file system or an http url.


src/common/localization/en-us.ts
```diff
@@ -231,6 +231,7 @@ export const english: IAppStrings = {
             nextAsset: "Next Asset",
             saveProject: "Save Project",
             exportProject: "Export Project",
+            activeLearning: "Auto Detect",
```
wbreza: Key does not match the label.

src/common/mockFactory.ts
```diff
@@ -283,6 +283,8 @@ export default class MockFactory {
             targetConnection: connection,
             tags: MockFactory.createTestTags(tagCount),
             videoSettings: MockFactory.createVideoSettings(),
+            activeLearningSettings: {modelPathType: "coco", modelPath: "", modelUrl: "",
```
wbreza: Nit: format whitespace.

src/electron/main.ts
```diff
@@ -10,7 +10,7 @@ import LocalFileSystem from "./providers/storage/localFileSystem";
 let mainWindow: BrowserWindow;
 let ipcMainProxy: IpcMainProxy;
 
-function createWindow() {
+async function createWindow() {
```
wbreza: Why making `createWindow` async? I don't see any calls to async methods.

src/models/applicationState.ts
```diff
@@ -198,6 +199,23 @@ export interface IProjectVideoSettings {
     frameExtractionRate: number;
 }
 
+/**
+ * @name - Video Tagging Settings for the project
```
wbreza: Docs are incorrect.  Also - If you are going to add `@member` declaration please also add docs for each member.

src/providers/activeLearning/electronProxyHandler.test.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as tf from "@tensorflow/tfjs";
+// tslint:disable-next-line:no-var-requires
+const modelJson = require("../../../cocoSSDModel/model.json");
+
+describe("Load default model from filesystem with TF io.IOHandler", () => {
+    it("Check file system proxy is correctly called", async () => {
+        const storageProviderMock = LocalFileSystemProxy as jest.Mock<LocalFileSystemProxy>;
+        storageProviderMock.mockClear();
+
+        storageProviderMock.prototype.readText = jest.fn((fileName) => {
```
wbreza: these methods are `async` and should therefore return a `Promise<T>`

src/providers/activeLearning/electronProxyHandler.test.ts
```diff
@@ -0,0 +1,33 @@
...
+            return JSON.stringify(modelJson);
+        });
+
+        storageProviderMock.prototype.readBinary = jest.fn((fileName) => {
+            return [];
+        });
+
+        const handler = new ElectronProxyHandler("folder");
+        try {
```
wbreza: Why empty catch block?  If the model fails it should bubble up or be handled.

src/providers/activeLearning/objectDetection.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import { ObjectDetection, DetectedObject } from "./objectDetection";
+// tslint:disable-next-line:no-var-requires
+const modelJson = require("../../../cocoSSDModel/model.json");
+
+describe("Load an Object Detection model", () => {
+    it("Load model from file system using proxy", async () => {
+        const storageProviderMock = LocalFileSystemProxy as jest.Mock<LocalFileSystemProxy>;
+        storageProviderMock.mockClear();
+
+        storageProviderMock.prototype.readText = jest.fn((fileName) => {
```
wbreza: Similar to above - these methods are async and should return a promise.

src/providers/activeLearning/objectDetection.ts
```diff
@@ -0,0 +1,195 @@
+import axios from "axios";
+import * as tf from "@tensorflow/tfjs";
+import { ElectronProxyHandler } from "./electronProxyHandler";
+import { LocalFileSystemProxy, ILocalFileSystemProxyOptions } from "../../providers/storage/localFileSystemProxy";
+
+// tslint:disable-next-line:interface-name
+export interface DetectedObject {
```
wbreza: Why are you not prefixing interface with `I` ?

src/providers/activeLearning/objectDetection.ts
```diff
@@ -0,0 +1,195 @@
...
+     * assocated class and score.
+     *
+     * @param img The image to detect objects from. Can be a tensor or a DOM
+     *     element image, video, or canvas.
+     * @param maxNumBoxes The maximum number of bounding boxes of detected
+     * objects. There can be multiple objects of the same class, but at different
+     * locations. Defaults to 20.
+     *
+     */
+    public async detect(img: tf.Tensor3D|ImageData|HTMLImageElement|HTMLCanvasElement|HTMLVideoElement,
```
wbreza: Consider declaring a typescript `type` that includes all these other type options.

src/providers/activeLearning/objectDetection.ts
```diff
@@ -0,0 +1,195 @@
...
+     * Dispose the tensors allocated by the model. You should call this when you
+     * are done with the model.
+     */
+    public dispose() {
+        if (this.model) {
+            this.model.dispose();
+        }
+    }
+
+    public async load(modelFolderPath: string) {
```
wbreza: Instead of having your `ObjectDetection` object handle the different types of loading a model - why don't we implement our own `IModelLoader` or `IModelProvider` which can be implemented to load a model as needed.  Then the `ObjectDetection` class can be simplified and just accept a `IModelProvider` during construction.

src/providers/activeLearning/objectDetection.ts
```diff
@@ -0,0 +1,195 @@
...
+        indexTensor.dispose();
+
+        // restore previous backend
+        tf.setBackend(prevBackend);
+
+        return this.buildDetectedObjects(width, height, boxes, maxScores, indexes, classes);
+    }
+
+    private buildDetectedObjects(
+        width: number, height: number, boxes: Float32Array, scores: number[],
```
wbreza: Lots of parameters here - consider refactoring this into an object.

src/react/components/common/assetPreview/imageAsset.tsx
```diff
@@ -12,7 +12,8 @@ export class ImageAsset extends React.Component<IAssetProps> {
             <img ref={this.image}
                 src={this.props.asset.path}
                 onLoad={this.onLoad}
-                onError={this.props.onError} />);
+                onError={this.props.onError}
+                crossOrigin="anonymous" />);
```
wbreza: What is this for?

src/react/components/pages/editorPage/canvas.tsx
```diff
@@ -205,26 +205,13 @@ export default class Canvas extends React.Component<ICanvasProps, ICanvasState>
...
-        this.deleteRegionsFromAsset(this.state.currentAsset.regions);
-    }
-
-    private addRegions = (regions: IRegion[]) => {
-        this.addRegionsToCanvasTools(regions);
-        this.addRegionsToAsset(regions);
-    }
-
-    private addRegionsToAsset = (regions: IRegion[]) => {
+    public addRegionsToAsset = (regions: IRegion[]) => {
```
wbreza: Instead of introducing new public methods to the canvas we should just expose `regions` as a new prop that appends the regions onto the canvas if they don't already exist.

src/react/components/pages/editorPage/editorPage.test.tsx
```diff
@@ -807,6 +818,73 @@ describe("Editor Page Component", () => {
...
+            const props = MockFactory.editorPageProps(testProject.id);
+            wrapper = createComponent(store, props);
+
+            editorPage = wrapper.find(EditorPage).childAt(0);
+            await waitForSelectedAsset(wrapper);
+            wrapper.update();
+
+            assetTestCache.clear();
+
+            document.createElement = jest.fn((elementType) => {
```
wbreza: Consider moving this whole block to the `MockFactory` and reusing it in the different places

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -46,6 +51,10 @@ export interface IEditorPageProps extends RouteComponentProps, React.Props<Edito
     applicationActions: IApplicationActions;
 }
 
+interface IExportPageSettings extends IAssetPreviewSettings {
```
wbreza: If this interface is going to exist as a prop on the page it needs to be exported and it should be moved up into `applicationState` with our application level interfaces.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -129,6 +142,38 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+                modelPath = this.props.project.activeLearningSettings.modelPath;
+            }
+        } else {
+            modelPath = this.props.project.activeLearningSettings.modelUrl;
+        }
+
+        // Load TensorFlow.js Model
+        this.model = new ObjectDetection();
+        const infoId = toast.info("Loading model...", { autoClose: false });
+        this.model.load(modelPath).then(() => {
```
wbreza: Please using `async` / `await` vs promise `then` callbacks

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -142,9 +187,17 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
                 project: this.props.project,
-                additionalSettings: { videoSettings: (this.props.project) ? this.props.project.videoSettings : null },
+                additionalSettings: { videoSettings: (this.props.project) ? this.props.project.videoSettings : null,
+                    activeLearningSettings: (this.props.project) ? this.props.project.activeLearningSettings : null },
             });
         }
+
+        if (this.state.project &&
+            this.state.project.activeLearningSettings.autoDetect &&
+            this.state.selectedAsset &&
```
wbreza: Consider encapsulating this into a variabled named `shouldPredictModel` or similar to make it easier to understand what is happening here.

We also should not be running predictions for every state change.  We should only predict when the selected asset has changed.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -252,6 +305,11 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         });
     }
 
+    private getAppPath = () => {
```
wbreza: This method is not related to the editor page and doesn't belong here.  Please move it out to a common place where it can be referenced if needed.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -484,9 +542,103 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
             case ToolbarItemName.RemoveAllRegions:
                 this.canvas.current.confirmRemoveAllRegions();
                 break;
+            case ToolbarItemName.ActiveLearning:
+                await this.predict();
+                break;
+        }
+    }
+
+    private predict = async () => {
```
wbreza: This code is very generic and should be moved out into a common location where you can pass in a `canvas` element and ultimately return an array of `IRegion` that can then get added to the canvas.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -484,9 +542,103 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+        await this.props.actions.saveProject(this.props.project);
+        // Temporary comment this as causing an issue with the dispatcher - to be debugged
+        // await this.updateProjectTagsFromAsset(newAsset, this.props.project, true);
+
+        this.setState({
+            selectedAsset: newAsset,
+        }, async () => {
+            await this.onAssetMetadataChanged(newAsset);
+        });
+    }
```
wbreza: The bulk of this method should be moved up into a common library.  I would expect to see it broken down into the following:

1. Method that takes a `canvas` input and returns an array of `regions`.
2. If project prediction is enabled, run prediction, set `regions` prop on canvas.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -541,16 +693,16 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         });
     }
 
-    private async updateProjectTagsFromAsset(asset: IAssetMetadata) {
+    private async updateProjectTagsFromAsset(asset: IAssetMetadata, project: IProject, forceSave: boolean) {
```
wbreza: Why are we passing in a project vs. using the current project?
In what scenario do we need to force save? Is it to save predicted regions?  If so - we should refactor to make this fit into the existing data flow.

src/react/components/pages/projectSettings/projectForm.json
```diff
@@ -41,6 +41,78 @@
                 "frameExtractionRate"
             ]
         },
+        "activeLearningSettings": {
```
wbreza: I would like to see this broken down into its own form similar to export settings.  Longer term we are going to update the UI to a tabbed interface where active learning deserves its own tab - but for now let's move out into separate form.

src/react/components/pages/projectSettings/projectForm.json
```diff
@@ -41,6 +41,78 @@
...
+            "title": "Active Learning ",
```
wbreza: Nit: Localize strings

src/electron/main.ts
JacopoMangiavacchi: Forgot to restore this after some testing with Electron.Remote

src/providers/activeLearning/electronProxyHandler.test.ts
JacopoMangiavacchi: We are not really loading the model here.  Weights are not provided/mocked so loading the model here will fail.  As in the description of the test the goal here is to test the ElectronProxyHandler class that implement the tensorflow.js io.IOHandler interface.  In particular here we want to check the passing a filesystem path the internal TF api call our proxied readText and readBinary methods to access the filesystem from the browser.  The readText is mocked to return a real tf.js json model with link to 5 binary weights file.  The weight files are not mocked but the test verify that the readBinary is actually called 5 times.

src/providers/activeLearning/objectDetection.ts
JacopoMangiavacchi: Good point.  It was like that in the original Google sample code.  From my point of view I think this is correct as this is not properly an Interface in the terms that this TS Interface is no where implemented in the code.  It is just a JS Object with no methods.  I think it's a kind of misleading of the TypeScript Interface feature.

JacopoMangiavacchi: Should this be eventually a TS Type vs an Interface ?

src/react/components/common/assetPreview/imageAsset.tsx
JacopoMangiavacchi: To allow to call canvas.getBlob in the browser

src/providers/activeLearning/objectDetection.ts
JacopoMangiavacchi: Not sure I see a real advantage here.  Let's talk about it.

src/providers/activeLearning/objectDetection.ts
JacopoMangiavacchi: I don't see the point here.  It's a method with 6 parameters.  Is the suggestion about performance, because of the difference of calling by value default type and byref object type, or is this about a better readable code ?

For performance reason I though that copy on write optimization in JS should completely manage this but I'm really eager to learn more here.

For code readability I don't see the point to create an object just to reduce the number of parameters between two private functions but I may be wrong.  It seams to me that wrapping all of this on an object is instead complicating the understanding but once again this is just MHO and eager to learn more.

src/providers/activeLearning/objectDetection.ts
```diff
@@ -0,0 +1,197 @@
...
+                }
+                const handler = new ElectronProxyHandler(modelFolderPath);
+                this.model = await tf.loadGraphModel(handler);
+
+                const provider = new LocalFileSystemProxy();
+                this.jsonClasses = JSON.parse(await provider.readText(modelFolderPath + "/classes.json"));
+            }
+
+            // Warmup the model.
+            const result = await this.model.executeAsync(tf.zeros([1, 300, 300, 3])) as
```
pjlittle: just curious about what 'warming up the model' - and the magic numbers in the array - actually do :slightly_smiling_face:

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: The call with force save is commented and it's the thing I need to show you on debug.

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: @PIC123 implemented this but I see we are not running prediction on every state change.  We are checking at every state change if automatic prediction is enabled and if the asset eventually has been already predicted once before.  Is there another event handler where to put this logic other then componentDidUpdate() ?

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: Do you have a suggestion what common file ?  applicationState.ts ??

src/providers/activeLearning/objectDetection.ts
```diff
@@ -0,0 +1,197 @@
...
+            });
+        }
+        return objects;
+    }
+
+    private getClass(index: number, indexes: Float32Array, classes: number[]): string {
+        if (index < indexes.length && indexes[index] < classes.length) {
+            const classId = classes[indexes[index]] - 1;
+            const classObject = this.jsonClasses[classId];
+            return classObject ? classObject.displayName : "";
```
pjlittle: would 'unknown' or 'not found' be more informative than an empty string? Is this displayed in the UI? Trying to think of how it would hit this case...

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -129,6 +138,36 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+            if (isElectron()) {
+                modelPath = this.props.project.activeLearningSettings.modelPath;
+            }
+        } else {
+            modelPath = this.props.project.activeLearningSettings.modelUrl;
+        }
+
+        // Load TensorFlow.js Model
+        this.model = new ObjectDetection();
+        const infoId = toast.info("Loading model...", { autoClose: false });
```
pjlittle: we should pull this string from the localization strings.

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -129,6 +138,36 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+            if (isElectron()) {
+                const appPath = this.getAppPath();
+
+                if (Env.get() !== "production") {
+                    modelPath = appPath + "/cocoSSDModel";
+                } else {
+                    modelPath = appPath + "/../../cocoSSDModel";
+                }
+            } else {
+                modelPath = "https://jmangiadiag.blob.core.windows.net/vottcontainer";
```
pjlittle: I don't have a great answer here, but we have to be careful about hosting things ourselves, since we can't guarantee it's always going to be there. 

A couple thoughts: 
- can/should we embed this in the SPA, version it just like any other asset
- Is there a better 'public' place for this to live?

src/react/components/pages/projectSettings/projectForm.json
JacopoMangiavacchi: Completely agree with you but just wondering if we are sure we want to do this in this PR ?  I mean this is not probably the only thing should be stripped out, think about tags for example.  Why we want to start with this and why in this PR ?

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -484,9 +536,103 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+                    const image64 = btoa(buffer.reduce((data, byte) => data + String.fromCharCode(byte), ""));
+                    const image = document.createElement("img");
+
+                    image.onload = async () => {
+                        this.predictImage(image);
+                    };
+                    image.src = "data:image;base64," + image64;
+                });
+            // tslint:disable-next-line:no-empty
+            } catch (_) {
```
pjlittle: just curious on the empty catch; does this fire if the image fails to load or ?  Would it make sense to log or is this an anticipated/deterministic failures?

src/providers/activeLearning/objectDetection.ts
JacopoMangiavacchi: :-)
The Tensor passed as input is a fake 300x300 image with 3 depth/colors.  The meaning for "warming up the model" is to fake a detection call in order to let the model load all necessary weights for his DL layers.

src/providers/activeLearning/objectDetection.ts
JacopoMangiavacchi: Good point.  Yes this is also UI related.  It is the tag name to use just in case there is a incongruence between the model and the json file with all tags managed by the model.  The orginal Google code for Coco image detection was simply embedding all this tags in code in a js object.  I extended this with this concept of a json file to dinamically pass a list of all the tags your model detect.  This is because of course the model doesn't use strings but it encode that in categorized label (numbers).  

Anyway probably you're right and "Unknown" is a much better label than an empty one!!!

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: My proposal in the mail was exactly to deploy this in the SPA but we need to talk about if we need to make some change to the pipeline.

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: In my test, specially with the WebApp, there are some strange issues where some of the internal calls my throw an exception in rare case.  For example even enabling the CORS server configuration and forcing crossOrigin="anonymous" in img/video object some times the canvas.getBlob just throw an exception.

IMHO auto detection (aka active learning) is an "utility" feature and in this particular case I don't think we should bored the user with an alert but we should simply do not detect nothing and warn no message at all.  But this is just my opinion here.  I don't know you but I hate windows like alert/messagebox and I prefer to use it only when strictly necessary.

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: Completely agree but there are many other cases like this.  For example the "Start/Finish exporting" toast and other.

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: Done but it receives an HTMLImageElement, not a canvas

src/react/components/pages/editorPage/editorPage.tsx
JacopoMangiavacchi: Partially done, see comments above.  Need to test new refactored method and definitely we need to debug a thing together on adding regions/tags to the project.

PR #741, opened 2019-04-11
fix: don't allow invalid path characters in project name

pjlittle:
Regex now prevents invalid path characters (<>"/\\|?*) in project
names, since it is used in file system operations. Using reserved
characters for Windows, as it is a superset of OSX/Linux.

[AB#18204](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=18204)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
         "rimraf": "^2.6.2",
         "shortid": "^2.2.14",
         "video-react": "^0.13.2",
-        "vott-ct": "^2.1.24",
+        "vott-ct": "2.1.24",
         "vott-react": "^0.2.10"
     },
     "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
         "rimraf": "^2.6.2",
         "shortid": "^2.2.14",
         "video-react": "^0.13.2",
-        "vott-ct": "^2.1.24",
+        "vott-ct": "2.1.24",
```
pjlittle: locking us in! :lock:

src/react/components/pages/projectSettings/projectForm.json
```diff
@@ -3,7 +3,8 @@
     "properties": {
         "name": {
             "title": "${strings.common.displayName}",
-            "type": "string"
+            "type": "string",
+            "pattern": "^[^\\\\\\\\/:*?\\\\\\\"<>|]*$"
```
pjlittle: Don't ask, don't tell... the trick was to escape this for both JSON and regex chars.  

Effectively won't allow a project name that contains: `<>"/|?*`

PR #742, opened 2019-04-11
fix: Update path to work on windows & refactor logic

tbarlow12:


src/common/utils.ts
```diff
@@ -51,6 +51,26 @@ export function createQueryString(object: any): string {
     return parts.join("&");
 }
 
+export function encodeFileURI(path: string, additionalEncodings?: boolean): string {
```
wbreza: Can you write some unit tests for these helper methods?

src/services/importService.ts
```diff
@@ -232,29 +225,50 @@ export default class ImportService implements IImportService {
...
+    private getAssetState(originalProject: any, frameRegions: IV1Region[], frameIntOrName: number|string): AssetState {
+        return originalProject.visitedFrames.indexOf(frameIntOrName) > -1 && frameRegions.length > 0
+        ? AssetState.Tagged : (originalProject.visitedFrames.indexOf(frameIntOrName) > -1
+        ? AssetState.Visited : AssetState.NotVisited);
+    }
+
+    private async getPopulatedAssetMetadata(
+            assetService: AssetService, asset: IAsset, assetState: AssetState,
+            frameRegions: IV1Region[], parent?: IAsset, includePath?: boolean): Promise<IAssetMetadata> {
+        return await assetService.getAssetMetadata(asset).then((metadata) => {
```
wbreza: Let's not mix `async` `await` with Promise `then` callback methods.  Please standardize to async/await.

src/services/importService.ts
```diff
@@ -200,21 +197,17 @@ export default class ImportService implements IImportService {
...
+            if (!frameList.hasOwnProperty(frameName)) {
+                continue;
             }
+            const frameRegions = frameList[frameName];
+            const asset = AssetService
+                .createAssetFromFilePath(`${v1Project.file.path.replace(/[^\/]*$/, "")}${frameName}`);
+            const assetState = this.getAssetState(originalProject, frameRegions, frameName);
+            generatedAssetMetadata.push(
+                await this.getPopulatedAssetMetadata(
+                    assetService, asset, assetState, frameRegions, undefined, true),
```
wbreza: `assetService` shouldn't need to be passed around.  We should be able to make it a instance member on the service.

PR #743, opened 2019-04-11
fix: Updated keyboard manager to support meta (aka cmd) key

wbreza:
Resolves an issue where normal Cmd+ keyboard shortcuts that typically work on Mac OSX don't work in VoTT.  The fix treats Ctrl or Cmd to be equivalent and also checks the value of the "meta" key within keyboard events.

Resolves [AB#18159](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=18159)


pjlittle:
Looks good.

PR #744, opened 2019-04-11
fix: Update tag names or delete tags from asset metadata [AB#17862]

tbarlow12:


src/services/projectService.ts
```diff
@@ -100,6 +103,48 @@ export default class ProjectService implements IProjectService {
...
+    private async updateProjectTags(
+            project: IProject, tag: string, currentAsset: IAssetMetadata, transformer: (tags: string[]) => string[]) {
+        const assetService = new AssetService(project);
+        const assetKeys = Object.keys(project.assets);
+        for (const assetKey of assetKeys) {
+            const asset = project.assets[assetKey];
+            const assetMetadata = await assetService.getAssetMetadata(asset);
+            const updatedAssetMetadata = this.updateAssetMetadata(assetMetadata, tag, transformer);
+            if (updatedAssetMetadata) {
+                assetService.save(updatedAssetMetadata);
```
wbreza: This `save` method is `async`

src/services/projectService.ts
```diff
@@ -100,6 +103,48 @@ export default class ProjectService implements IProjectService {
...
+                           newTag: string, currentAsset: IAssetMetadata): Promise<IAssetMetadata> {
+        return await this.updateProjectTags(
+            project, tag, currentAsset, (tags) => tags.map((t) => (t === tag) ? newTag : t));
+    }
+
+    private async updateProjectTags(
+            project: IProject, tag: string, currentAsset: IAssetMetadata, transformer: (tags: string[]) => string[]) {
+        const assetService = new AssetService(project);
+        const assetKeys = Object.keys(project.assets);
+        for (const assetKey of assetKeys) {
```
wbreza: Let's wrap this loop in a `foreEachAsync`

src/services/projectService.ts
```diff
@@ -100,6 +103,48 @@ export default class ProjectService implements IProjectService {
...
+    public async deleteTag(project: IProject,
+                           tag: string, currentAsset: IAssetMetadata): Promise<IAssetMetadata> {
+        return await this.updateProjectTags(
+            project, tag, currentAsset, (tags) => tags.filter((t) => t !== tag));
+    }
+
+    public async updateTag(project: IProject, tag: string,
+                           newTag: string, currentAsset: IAssetMetadata): Promise<IAssetMetadata> {
+        return await this.updateProjectTags(
+            project, tag, currentAsset, (tags) => tags.map((t) => (t === tag) ? newTag : t));
```
wbreza: I would break out the transformer into separate variable to make it easier to understand what is going on.

src/services/projectService.ts
```diff
@@ -100,6 +103,48 @@ export default class ProjectService implements IProjectService {
...
+    }
+
+    private async updateProjectTags(
```
wbreza: Let's add some docs here to make sure it is clear what the transformers purpose is.

src/services/projectService.ts
wbreza: I think we can be a bit more proactive here and only update the assets that are currently in the `Tagged` state.

wbreza: There could also be a use case where if all regions are now untagged we need to remove the region which should cause the asset state to transition back into the `Visited` state.

src/services/projectService.test.ts
```diff
@@ -146,15 +148,106 @@ describe("Project Service", () => {
...
+        };
+        const assetMetadata = MockFactory.createTestAssetMetadata(asset, [region]);
+        AssetService.prototype.getAssetMetadata = jest.fn((asset: IAsset) => Promise.resolve(assetMetadata));
+
+        const saveMetadata = jest.fn();
+        AssetService.prototype.save = saveMetadata;
+
+        const expectedAssetMetadata: IAssetMetadata = MockFactory.createTestAssetMetadata(asset, []);
+        const project = populateProjectAssets();
+        await projectSerivce.deleteTag(project, tag1);
```
wbreza: Let's also check that any assets without tagged regions transition back to `Visited` state.

src/react/components/common/tagInput/tagInput.tsx
```diff
@@ -31,9 +31,9 @@ export interface ITagInputProps {
     /** Function to call on clicking individual tag while holding CTRL key */
     onCtrlTagClick?: (tag: ITag) => void;
     /** Function to call when tag is renamed */
-    onTagRenamed?: (oldTag: string, newTag: string) => void;
+    onTagRenamed?: (tag: string, newTag: string) => void;
```
mydiemho: I would rename these to `tagName`

src/react/components/common/tagInput/tagInput.tsx
```diff
@@ -31,9 +31,9 @@ export interface ITagInputProps {
...
     /** Function to call when tag is deleted */
-    onTagDeleted?: (tag: ITag) => void;
+    onTagDeleted?: (tag: string) => void;
```
mydiemho: same `tagName`

src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -436,6 +439,24 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
         });
     }
 
+    private onTagDeleted = async (tag: string) => {
+        const projectService = new ProjectService();
+        const updatedAsset = await projectService.deleteTag(this.props.project, tag, this.state.selectedAsset);
+        if (updatedAsset) {
```
mydiemho: if `updatedAsset` is false, does that mean delete operation failed?

src/services/projectService.ts
```diff
@@ -18,6 +19,8 @@ export interface IProjectService {
     save(project: IProject, securityToken: ISecurityToken): Promise<IProject>;
     delete(project: IProject): Promise<void>;
     isDuplicate(project: IProject, projectList: IProject[]): boolean;
+    deleteTag(project: IProject, tag: string, currentAsset: IAssetMetadata): Promise<IAssetMetadata>;
```
mydiemho: `tagName`

src/services/projectService.ts
```diff
@@ -18,6 +19,8 @@ export interface IProjectService {
...
+    updateTag(project: IProject, tag: string, newTag: string, currentAsset: IAssetMetadata): Promise<IAssetMetadata>;
```
mydiemho: `tagName`

src/services/projectService.test.ts
```diff
@@ -146,15 +148,106 @@ describe("Project Service", () => {
...
+                ],
+            ),
+
+        };
+        const project = populateProjectAssets();
+        await projectSerivce.deleteTag(project, tag1, assetMetadata);
+        expect(saveMetadata).toBeCalledWith(expectedAssetMetadata);
     });
+
+    it("Deletes any empty regions after deleting only tag from region", async () => {
```
mydiemho: I think this is very aggressive and we should alert user warning them if they're deleting the only tag of a region

tbarlow12:
Closing in favor of newer PR

PR #748, opened 2019-04-12
doc: update v1/master reference

pjlittle:
Update release script to reference master, instead of v2.


scripts/release-pr.sh
```diff
@@ -42,7 +42,7 @@ then
     echo "version=${NEW_VERSION}"
 
     # get the latest from v2, create a release branch
-    git checkout v2
+    git checkout master
```
mydiemho: I included this change in my PR

PR #750, opened 2019-04-13
Update release-pipeline.yml for Azure Pipelines

mydiemho:


release-pipeline.yml
```diff
@@ -1,5 +1,5 @@
 trigger:
-  - refs/tags/master # only run build on tag with this pattern
+  - refs/tags/v2* # only run build on tag with this pattern
```
pjlittle: do we want to just make this: `refs/tags/*` so that any tag will build artifacts?

mydiemho: I'm fine with that, the release has a condition to only release on master so we're safe

PR #757, opened 2019-04-14
feat: Add CSV Exporter

JacopoMangiavacchi:
Adds CSV export provider to VoTT utilizing `json2csv` npm package

- Options
  - Asset State
  - Include Images


src/providers/export/csvFormat.ts
```diff
@@ -0,0 +1,62 @@
...
+                // Push CSV Records
+                // The CSV file itself must have the following format::
+                // image_id,xmin,ymin,xmax,ymax,label
+                // image_1.jpg,26,594,86,617,cat
+                // image_1.jpg,599,528,612,541,car
+                // image_2.jpg,393,477,430,552,dog
+                assetMetadata.regions.forEach((region) => {
+                    region.tags.forEach((tag) => {
+                        csvBuffer.push([
+                            assetMetadata.asset.name,
```
pjlittle: I think we're safe on the numbers, but should we wrap each field in double quotes, or otherwise leverage a CSV writer, in case someone has an asset or tag name that would break the format?

pjlittle:
looks good, just have one question. At some point we may want to use a file/CSV writer so that we don't buffer everything in memory, but for now should be fine.

SonarCloud is complaining about duplicate lines, but it's unclear to me as to why.

src/providers/export/csvFormat.ts
JacopoMangiavacchi: Good point as actually comma is accepted in file name.   I didn't see wrapping file on quotes on the examples I found but it should work.

JacopoMangiavacchi:
@wbreza unfortunately I started this from the Active Learning branch so if I re-target to develop it will contains all the commits from that branch too.

I think we have to first approve the AL PR.  If this is urgent of course we could just copy the changes manually on a new branch but is this really so urgent ?

Issue #760, opened 2019-04-15
PascalVOC export generating wrong value for ymax


DaiSatoJ:
When i try to export TFPascalVOC format, ymax is identical to xmin in bounding box value.

exported xml
```
<annotation verified="yes">
    <folder>Annotation</folder>
    <filename>A_001.mp4#t=0.jpg</filename>
    <path>A-TFPascalVOC-export/Annotations/A_001.mp4#t=0.jpg</path>
    <source>
        <database>Unknown</database>
    </source>
    <size>
        <width>1280</width>
        <height>720</height>
        <depth>3</depth>
    </size>
    <segmented>0</segmented>
    <object>
    <name>object</name>
    <pose>Unspecified</pose>
    <truncated>0</truncated>
    <difficult>0</difficult>
    <bndbox>
        <xmin>292.3342939481268</xmin>
        <ymin>468.6683738796415</ymin>
        <xmax>723.9193083573487</xmax>
        <ymax>292.3342939481268</ymax>
    </bndbox>
</object>
</annotation>
```

and corresponding asset json file
```
{
    "asset": {
        "id": "fd5c55428d521d2419d7d75aa23847ef",
        "format": "mp4",
        "state": 2,
        "type": 3,
        "name": "A_001.mp4#t=0",
        "path": "file:F:/dataset/B/video/A/A_001.mp4#t=0",
        "size": {
            "width": 1280,
            "height": 720
        },
        "parent": {
            "format": "mp4",
            "id": "3c210bd8fb5d485fc70ad3e2683a5fc1",
            "name": "A_001.mp4",
            "path": "file:F:/dataset/B/video/A/A_001.mp4",
            "size": {
                "width": 1280,
                "height": 720
            },
            "state": 1,
            "type": 2
        },
        "timestamp": 0
    },
    "regions": [
        {
            "id": "Bq6Qilz7A",
            "type": "RECTANGLE",
            "tags": [
                "object"
            ],
            "boundingBox": {
                "height": 126.29961587708067,
                "width": 431.5850144092219,
                "left": 292.3342939481268,
                "top": 468.6683738796415
            },
            "points": [
                {
                    "x": 292.3342939481268,
                    "y": 468.6683738796415
                },
                {
                    "x": 723.9193083573487,
                    "y": 468.6683738796415
                },
                {
                    "x": 723.9193083573487,
                    "y": 594.9679897567222
                },
                {
                    "x": 292.3342939481268,
                    "y": 594.9679897567222
[...]

PR #763, opened 2019-04-15
fix: Correctly save the ymax value on PascalVOC exporter and remove Tensorflow name

JacopoMangiavacchi:
- fix issue #760 [[AB#18223](https://dev.azure.com/dwrdev/web/wi.aspx?pcguid=80f9922c-b416-454c-92c6-ab7b6867e0b2&id=18223)]


pjlittle:
glad we got the name straightened out :slightly_smiling_face:

PR #764, opened 2019-04-15
fix: Rename/delete tags throughout assets [AB#17862]

tbarlow12:


src/react/components/pages/editorPage/editorPage.tsx
```diff
@@ -288,6 +304,69 @@ export default class EditorPage extends React.Component<IEditorPageProps, IEdito
...
+     */
+    private onTagRenamed = async (tagName: string, newTagName: string): Promise<void> => {
+        const { project, selectedAsset } = this.state;
+        const newProject = {
+            ...project,
+            tags: project.tags.map((t) => (t.name === tagName) ? {...t, name: newTagName} : t),
+        };
+
+        const assetService = new AssetService(newProject);
+        const asset = await assetService.renameTag(project.assets, tagName, newTagName, selectedAsset);
```
wbreza: `AssetService` already has a reference to a project (which has the assets).  Any reason to still pass the assets here?

wbreza: Also seems a bit strange that this takes / and returns the selected asset?

tbarlow12: 1. Kinda forgot about that. Good call
2. I was going to ask you about that. I was just looking for a good way to refresh the current selected asset. It seemed like it would be less efficient to try and load it from the file again since I was going to use it right there. Open to suggestions

PR #765, opened 2019-04-15
Fix: Enables selection of azure region for custom vision export

### LINKED ISSUE #759: add support other regions of custom vision service
please add support for other regions of custom vision service apart from south central us,
currently it's hard-coded and it's not that informative...
I've discovered that only by enabling developer mode.

wbreza:
Adds ability to select azure region where your custom vision service is hosted.
Filters domain list by project type

Resolves #759, #770


src/providers/export/azureCustomVision.ts
```diff
@@ -38,6 +39,24 @@ export enum NewOrExisting {
     Existing = "existing",
 }
 
+/**
+ * Azure regions
+ */
+export enum AzureRegion {
```
pjlittle: we're not using elsewhere, but I can see this bubbling up at some point, outside of CV.

wbreza: Agreed - will move it up at that time.

PR #782, opened 2019-04-25
fix: Fix display of tag color picker

tbarlow12:
- Updates the `TagInput` component to use the `TagInputItem` type instead of `RefObject<TagInputItem>`
- Adds several unit tests of uncovered areas for `TagInput`


tbarlow12:
Resolves [AB#18368]

PR #785, opened 2019-04-26
feat: Hash asset IDs with project name [AB#18063]

tbarlow12:
- Hashes asset file IDs with project name to allow for multiple projects using the same assets
- Updates `IAsset` to contain `projectName` property
- Updates assetProvider to use `projectName` in the `getAssets` call

Resolves [[AB#18063](https://dev.azure.com/dwrdev/e2e1d5cd-80d1-4aa1-9b7d-b19c4bcaa288/_workitems/edit/18063)]

### DIFF
--- a/src/providers/storage/assetProviderFactory.ts
+++ b/src/providers/storage/assetProviderFactory.ts
@@ -9,7 +9,7 @@ import getHostProcess, { HostProcessType } from "../../common/hostProcess";
  */
 export interface IAssetProvider {
     initialize?(): Promise<void>;
-    getAssets(containerName?: string): Promise<IAsset[]>;
+    getAssets(projectName: string): Promise<IAsset[]>;
 }
 
 /**
--- a/src/services/assetService.ts
+++ b/src/services/assetService.ts
@@ -40,10 +40,10 @@ export class AssetService {
             filePath = encodeFileURI(filePath, true);
         }
 
-        const hashTarget = projectName ? JSON.stringify({
+        const hashTarget = JSON.stringify({
             projectName,
             filePath
-        }) : filePath
+        });
 
         const md5Hash = new MD5().update(hashTarget).digest("hex");
         const pathParts = filePath.split(/[\\\/]/);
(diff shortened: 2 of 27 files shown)


src/providers/storage/assetProviderFactory.ts
```diff
@@ -9,7 +9,7 @@ import getHostProcess, { HostProcessType } from "../../common/hostProcess";
  */
 export interface IAssetProvider {
     initialize?(): Promise<void>;
-    getAssets(containerName?: string): Promise<IAsset[]>;
```
tbarlow12: This `containerName` variable is never used in any of the asset providers. Open to putting it back in, but seemed unnecessary

src/services/assetService.ts
```diff
@@ -40,7 +40,12 @@ export class AssetService {
             filePath = encodeFileURI(filePath, true);
         }
 
-        const md5Hash = new MD5().update(filePath).digest("hex");
+        const hashTarget = JSON.stringify({
```
tbarlow12: This is just one way to create the string. I wasn't sure if we'd ever need to "go the other way" (reverse the hash and parse out the project file & file path). We can also just concatenate the strings. Open to feedback

src/services/assetService.ts
```diff
@@ -130,8 +136,8 @@ export class AssetService {
     /**
      * Get assets from provider
      */
-    public async getAssets(): Promise<IAsset[]> {
-        return await this.assetProvider.getAssets();
+    public async getAssets(projectName: string): Promise<IAsset[]> {
```
wbreza: The assetProvider already has a reference to the project so we shouldn't need to pass in the project name as a function argument.

wbreza:
Do we think this is a common use case?  This solves the use case of having 2 project referencing the same source assets and storing both projects in the same target destination... I can see how this might be easy to repro in development but is it a realistic use case?  What happens if the project is renamed after assets are added?

tbarlow12:
@wbreza I don't know for sure if this is a common use case or not. I just had some extra time and picked the next feature off the backlog that made sense for me to take on in a short amount of time.

The renaming of the project is something that I didn't account for. We can chat more about that or we can just abandon this and close that feature.

PR #789, opened 2019-04-29
fix: CNTK export images bug fix

wbreza:
Fixes an issue where the images exported out of a video file were missing file extension for video projects.

### DIFF
--- a/src/providers/export/cntk.ts
+++ b/src/providers/export/cntk.ts
@@ -76,12 +76,11 @@ export class CntkExportProvider extends ExportProvider<ICntkExportProviderOption
             });
         });
 
+        const buffer = await HtmlFileReader.getAssetArray(assetMetadata.asset);
         const folderName = exportSplit === ExportSplit.Train ? "positive" : "testImages";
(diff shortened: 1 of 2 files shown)


src/services/projectService.ts
```diff
@@ -171,7 +171,9 @@ export default class ProjectService implements IProjectService {
      * @param project The project to update
      */
     private ensureBackwardsCompatibility(project: IProject) {
-        if (project.version === "2.0.0") {
+        const projectVersion = project.version.toLowerCase();
```
wbreza: Handles case where user was using older `-preview` versions.

src/providers/export/cntk.ts
```diff
@@ -76,13 +76,12 @@ export class CntkExportProvider extends ExportProvider<ICntkExportProviderOption
             });
         });
 
+        const buffer = await HtmlFileReader.getAssetArray(assetMetadata.asset);
```
wbreza: The `HtmlFileReader` updates the incoming asset and appends a .jpg to the name.  Not the best practice but it's consistent with the other exporters.

pjlittle:
:+1:

PR #811, opened 2019-05-28
Update active learning classes.json processor

csaroff:
Previously, the code that loads in the classes.json file did not account for the missing classes in COCO.  This fixes associated indexing issues so that active learning with the default cocossd model works out of the box on the coco dataset.


tbarlow12:
@csaroff please re-target your PR to the `develop` branch

PR #823, opened 2019-06-09
fix: test asset distribution to include all tags on test/train split

hermanho:
The test asset may not included all tags when export with test/train split option and uneven distribution of tags in current venison (2.1.0). Here is an example:

Image 1-8 are tagged with Tag-1.
Image 9-10 are tagged with Tag-2.
Export provider: PascalVOC
Asset State: Only tagged Assets
Test / Train Split is 80%.

Output
tag-1_val.txt
```
asset-9.jpg -1
asset-10.jpg -1
```
tag-2_val.txt
```
asset-9.jpg 1
asset-10.jpg 1
```

In current venison, there is no images of tag "Tag-1" in the text file "tag-1_val.txt" because the split method is based on the whole image set so only image 9 & 10 are put in test asset. The split method is changed to extract the test assets by each tag and combine them at the end. Image 7,8,10 will put in test asset after the change applied. 

Output
tag-1_val.txt
```
asset-7.jpg 1
asset-8.jpg 1
asset-10.jpg -1
```
tag-2_val.txt
```
asset-7.jpg -1
asset-8.jpg -1
asset-10.jpg 1
```

### DIFF
--- a/src/providers/export/cntk.ts
+++ b/src/providers/export/cntk.ts
@@ -3,7 +3,7 @@ import { ExportProvider, IExportResults } from "./exportProvider";
 import { IAssetMetadata, IExportProviderOptions, IProject } from "../../models/applicationState";
 import HtmlFileReader from "../../common/htmlFileReader";
 import Guard from "../../common/guard";
-import {splitTestAsset} from "./testAssetsSplitHelper";
+import { splitTestAsset } from "./testAssetsSplitHelper";
 
 enum ExportSplit {
     Test,
--- a/src/providers/export/cntk.ts
+++ b/src/providers/export/cntk.ts
@@ -3,6 +3,7 @@ import { ExportProvider, IExportResults } from "./exportProvider";
 import { IAssetMetadata, IExportProviderOptions, IProject } from "../../models/applicationState";
 import HtmlFileReader from "../../common/htmlFileReader";
 import Guard from "../../common/guard";
+import {splitTestAsset} from "./testAssetsSplitHelper";
 
 enum ExportSplit {
     Test,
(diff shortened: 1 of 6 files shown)


tbarlow12:
@hermanho please re-target your PR to the `develop` branch

hermanho:
@tbarlow12 changed

src/providers/export/cntk.ts
```diff
@@ -3,6 +3,7 @@ import { ExportProvider, IExportResults } from "./exportProvider";
 import { IAssetMetadata, IExportProviderOptions, IProject } from "../../models/applicationState";
 import HtmlFileReader from "../../common/htmlFileReader";
 import Guard from "../../common/guard";
+import {splitTestAsset} from "./testAssetsSplitHelper";
```
wbreza: Nit: Spacing around `{` in import statements.

src/providers/export/testAssetsSplitHelper.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * A helper function to split train and test assets
+ * @param template String containing variables
+ * @param params Params containing substitution values
+ */
+export function splitTestAsset(allAssets: IAssetMetadata[], tags: ITag[], testSplitRatio: number): string[] {
+    const testAssets: string[] = [];
+
+    if (testSplitRatio > 0 && testSplitRatio <= 1) {
```
wbreza: Consider inverting if statement to reduce code nesting.

src/providers/export/cntk.ts
hermanho: Fixed in bfce1ec

src/providers/export/testAssetsSplitHelper.ts
hermanho: Fixed in 74a5a81

PR #842, opened 2019-07-29
feat: changed sidebar to be horizontal

riagarg:
Moved the vertical asset previewer to be horizontal and at the bottom of the screen.


src/registerToolbar.ts
```diff
@@ -142,7 +142,7 @@ export default function registerToolbar() {
     ToolbarItemFactory.register({
         name: ToolbarItemName.ExportProject,
         tooltip: strings.editorPage.toolbar.exportProject,
-        icon: "fa-external-link-square-alt",
+        icon: "fas fa-file-export",
```
elizabethhalper: This should be the same as it originally had been until your other feature gets merged in. We'd like to keep the features separate, so this should match with the current master for now.

src/react/components/pages/editorPage/editorSideBar.tsx
```diff
@@ -102,8 +100,8 @@ export default class EditorSideBar extends React.Component<IEditorSideBarProps,
         this.props.onAssetSelected(asset);
     }
 
-    private rowRenderer = ({ key, index, style }): JSX.Element => {
-        const asset = this.props.assets[index];
+    private rowRenderer = ({ columnIndex, key, rowIndex, style }): JSX.Element => {
```
elizabethhalper: It doesn't look like rowIndex is being used in this method, so you can get rid of it as a parameter.

src/react/components/pages/editorPage/editorSideBar.test.tsx
```diff
@@ -117,8 +117,8 @@ describe("Editor SideBar", () => {
         };
 
         const wrapper = createComponent(props);
-        const list = wrapper.find(List).instance() as List;
-        const recomputeRowHeightsSpy = jest.spyOn(list, "recomputeRowHeights");
+        const grid = wrapper.find(Grid).instance() as Grid;
+        const recomputeRowHeightsSpy = jest.spyOn(grid, "recomputeGridSize");
```
elizabethhalper: nit: to keep your method names consistent, this should be renamed "recomputeGridSizeSpy"

PR #871, opened 2019-08-21
feat: VoTT JSON File Schema Versioning

### LINKED ISSUE #814: Feature Request: JSON Schema Versioning
Hello,

I would like to have a field in the VOTT JSON files (*.vott, *-assets.json) that contains a schema revision/version identifier. We are building out our tooling and would like to be able to inspect the version to ensure that our tools will run knowing that the schema has not been altered. Currently the JSON stores the software version, but by the looks of it, the software updates are not making changes to the schema, or the changes to the schema may be out of step with the software version changes.

tbarlow12:
- [x] Added schema version to both `.vott` file and `-asset.json` files
- [x] Added constants to specify version for both
- [x] Did *not* try to add versions to existing projects. If not included, assume `<=1.0.0`
- [x] Added `fileExists` function to `StorageProvider` interface and implemented in all providers
- [x] Added unit tests to verify all functionality

Resolves #814

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -5,4 +5,6 @@ export const constants = {
     projectFileExtension: ".vott",
     assetMetadataFileExtension: "-asset.json",
     exportFileExtension: "-export.json",
+    projectSchemaVersion: "1.0.0",
+    assetSchemaVersion: "1.0.0",
 };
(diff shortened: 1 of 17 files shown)


src/common/constants.ts
```diff
@@ -5,4 +5,6 @@ export const constants = {
     projectFileExtension: ".vott",
     assetMetadataFileExtension: "-asset.json",
     exportFileExtension: "-export.json",
+    projectSchemaVersion: "1.0.0",
```
wbreza: The plan is to version the schema separately from the NPM project version?

wbreza: Thinking this version number should come from a separate package (vott-react?).  Otherwise it will be tricky to keep these in sync or to know when to update this schema version.

PR #872, opened 2019-08-21
chore: install dependabot

mydiemho:
Dependabot will check for dependencies update weekly

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,10 @@
+version: 1
+
+update_configs:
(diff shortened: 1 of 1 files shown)


pjlittle:
My only worry is that it will update dependencies, but we won't have a mechanism to review or test for breaking changes. Does this only do minor/patch updates?

We're finding tons of issues updating these right now.

mydiemho:
The bot will not commit directly to master/develop.  It will only create PR, which will require the changes passing all tests and code reviews.

PR #901, opened 2019-09-20
67 deploy vott on a server

ebreton:
This PR defines:

- `Dockerfile`s files for both dev (localhost) and remote server 
- `docker-compose...yml` which simplify running the images, and allow deployment in a docker swarm

All commands are compiled in a `Makefile` and the `README` is updated with a bit of explanation.

The deployment as been executed on https://vott-dev.cortexia.io and https://vott-qa.cortexia.io with the code source from this PR


ebreton:
sorry, needs to go to our repo first ...

PR #908, opened 2019-10-04
fix: Update plato report links

tbarlow12:


mydiemho:
@tbarlow12 I think you have already updated master directly

PR #945, opened 2020-03-06
Upgrade windows & mac agents per AzDO removing older support

mydiemho:


azure-pipelines/plato/generate-up-to-date-report.yml
```diff
@@ -51,7 +51,7 @@ steps:
 - task: AzureCLI@1
   displayName: "Pull down old report and add updates"
   inputs:
-    azureSubscription: 'PELITTLE TEAM - CSE DWR (d36d0808-a967-4f73-9fdc-32ea232fc81d)'
+    azureSubscription: 'pj-little-sub'
```
pjlittle: Good catch.

pjlittle:
Thanks for addressing this. I'm not sure we would necessarily want to specify 'latest' for images, but given that we're using specific versions like `windows-2019` what's a good strategy for keeping these up-to-date? Seems like it will inevitable break at some point. Is there anything else we can do?

azure-pipelines/plato/generate-up-to-date-report.yml
mydiemho: AzDO accepts service connection name as value so it's cleaner to use the name vs the long subscription id & name

mydiemho:
> Thanks for addressing this. I'm not sure we would necessarily want to specify 'latest' for images, but given that we're using specific versions like `windows-2019` what's a good strategy for keeping these up-to-date? Seems like it will inevitable break at some point. Is there anything else we can do?

There isn't really a way to keep them up-to-date per se.  We just have to check in periodically.  

Fortunately, AzDO has also been really good about sending emails about non-backward compatible deprecation.  For example, they block all pipelines with old agents from running so we know this has to be updated.

PR #964, opened 2020-05-18
chore: update & document release process

mydiemho:
swimlanes-9edfa565b100bdc65903c210b69a2eaa

GitHubRelease


azure-pipelines/create-release.yml
```diff
@@ -0,0 +1,33 @@
+name: "GitHub & Web Release [$(SourceBranchName)] - $(Date:yyyyMMdd)$(Rev:.r)"
```
wbreza, line 1: Was the old process not working?  Looks like a full rewrite on the pipelines?

azure-pipelines/templates/create-web-release.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - task: AzureFileCopy@3
+        displayName: 'AzureBlob File Copy to $DEV_STORAGE_ACCOUNT'
+        inputs:
+          SourcePath: './build'
+          azureSubscription: $(azureSubscription)
+          destination: azureBlob
+          storage: $(DEV_STORAGE_ACCOUNT)
+          containerName: '$web'
+        # condition: eq(variables['Build.SourceBranch'], 'refs/heads/develop')
```
wbreza, line 36: Should this line be commented out?

azure-pipelines/create-release.yml
mydiemho: a lot of the old process was manual.  A few of AzDO task has been updated to support the functionality we need so I removed the manual pieces.

azure-pipelines/templates/create-web-release.yml
mydiemho: good catch

Issue #970, opened 2020-06-01
Can VoTT export the labels in COCO format？


songyuc:
Hi guys,
I am a learner of CV (Computer Vision).
And I want to know whether VoTT can export the labels in COCO format?

Your answer or idea will be appreciated!

jeongjinhwi:
No, It can't....
So, I exported with pascal voc and used open source code.
I used this code : https://github.com/yukkyo/voc2coco/blob/master/voc2coco.py

songyuc:
Hi @jeongjinhwi , 
thanks sincerely for the sharing!

Issue #913, opened 2019-10-17
Import pre labeled vott-json or PascalVOC files


radusl:
How can i import pre labeled VOTT-json or PascalVOC files? There seems to be no way to do that. Is it even possible?

bhaskar2443053:
I also want this feature

mozpp:
I also want this feature
俺也一样

smilexin:
me too

fan-zhang-echonous:
Also in my wish-list:D

steinmann:
This would be great!

Chappie74:
Hey guys. I was looking for the same feature today and what I did was just created a script to convert my original pascal voc labeled data into the form that VOTT understands. Here is the repo. I hope it helps you all.  https://github.com/Chappie74/convert_pascalvoc_to_vott

Issue #902, opened 2019-09-24
Load Existing Labels 


mmortazavi:
**Is your feature request related to a problem? Please describe.**
As far as I played with the VoTT I could not load existing project (directory) where I have already images with labels (either xml or yolo). This allows to keep track of an existing folder when new images are added or even allow editing existing labels. 

**Describe the solution you'd like**
A feature, at least on the local version where access to local file-system is possible, that can also read and load the existing labels automatically would be nice considering both image and label file are in the same folder (or alternatively providing path to image and label folders explicitly).

**Describe alternatives you've considered**
There are a few tools out there that can offer such feature like https://github.com/tzutalin/labelImg, however, I liked to try VoTT since it seems to allows exporting to various formats easier and have a nicer interface for image tagging and labeling. 

**Additional context**
I am using Windows 10 version from the last release -> vott-2.1.0-win32.exe.

Issue #803, opened 2019-05-15
No Yolo Export Provider


xzessmedia:
In the description it says export to yolo but there is literally no information on that topic anywhere.
all git issues with the topic "yolo" has been closed.. are those closed issues fixed? it does not seem to.
why are those issues closed without any information about the result?

so please tell me how to export vott data to darknet / yolo ?

tbarlow12:
@xzessmedia I have removed the yolo tag for now, as we do not yet support it in v2. The [v1 branch](https://github.com/microsoft/VoTT/tree/v1) does support it, but that is no longer being maintained.

xzessmedia:
Thanks for this info,
right now i switched to a really simple piece of software which works just great "Boobs".
Maybe you can orient on that and it helps to enhance VoTT which has good potential.
It is much simpler, works far better right now for my purpose.

https://github.com/drainingsun/boobs

pullmyleg:
Here is a pyhton script to convert from VoTT CSV to Yolo label txt files. 

```
from PIL import Image
from os import path, makedirs
import os
import re
import pandas as pd
import sys
import argparse


def get_parent_dir(n=1):
    """returns the n-th parent dicrectory of the current
    working directory"""
    current_path = os.path.dirname(os.path.abspath(__file__))
    for k in range(n):
        current_path = os.path.dirname(current_path)
    return current_path


sys.path.append(os.path.join(get_parent_dir(1), "Utils"))
from Convert_Format import convert_vott_csv_to_yolo

Data_Folder = os.path.join(get_parent_dir(1), "Data")
VoTT_Folder = os.path.join(
    Data_Folder, "Source_Images", "Training_Images", "vott-csv-export"
)
VoTT_csv = os.path.join(VoTT_Folder, "Annotations-export.csv")
YOLO_filename = os.path.join(VoTT_Folder, "data_train.txt")

model_folder = os.path.join(Data_Folder, "Model_Weights")
classes_filename = os.path.join(model_folder, "data_classes.txt")

if __name__ == "__main__":
    # surpress any inhereted default values
    parser = argparse.ArgumentParser(argument_default=argparse.SUPPRESS)
    """
    Command line options
    """
    parser.add_argument(
        "--VoTT_Folder",
        type=str,
        default=VoTT_Folder,
        help="Absolute path to the exported files from the image tagging step with VoTT. Default is "
        + VoTT_Folder,
    )

    parser.add_argument(
        "--VoTT_csv",
        type=str,
[...]

PR #1002, opened 2020-09-25
docs: reference external tools for converting labels

### LINKED ISSUE #994: When does VOTT v2 support yolo format?
I really like this project, but it doesn't support exporting to yolo format. This makes me sad.There are a lot of people using yolo and I hope it helps those people!

yeldarby:
Add links in the docs to external tools to convert the VoTT schema to and from other formats.

VoTT does not yet support exporting to other popular formats (eg YOLO TXT) or importing labels from other annotation tools. There have been been several issues requesting this additional functionality.

fix #994 #970 #913 #902 #803 #800 #292 #262 #151 #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,13 @@ Click on the `Export` (arrow) icon in the left hand navigation. Select the appro
 
 ![alt text](docs/images/export-labels.jpg "Export Labels")
 
+### Converting Labels
+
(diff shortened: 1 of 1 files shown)


PIC123 (approved):
Looks good, thanks for the addition!

PR #1018, opened 2020-11-02
fix: Updates node-sass version to support local development

wbreza:
The current version of node-sass is very outdated which is causing build errors due to 404 not found requests to components no longer available.

This PR updates to latest version of node sass and confirms local development can be setup.


PIC123 (approved):
LGTM!

jding05 (approved):
LGTM

Issue #835, opened 2019-07-23
[Feature] Make Security Tokens Optional


MichaelCVelez:
**Is your feature request related to a problem? Please describe.**
If a project does not contain sensitive information, it is forced to be encrypted anyway and requires a matching token to be opened. This becomes an obstacle when trying to share projects.

**Describe the solution you'd like**
I would like to be able to turn token generation off before creating a project to make it accessible to others regardless of which tokens they have. Ideally you can also switch it back on before creating another project at any time.

**Describe alternatives you've considered**
I need to view several users projects and each have generated a unique token. I could request these tokens from them but they work on rotating schedules and could take a full week to get those tokens from them. I could have them create one token that is shared between all projects, but they could forget to use that token on project creation. 

**Additional context**
Since this is an internal group, we should be able to open each other’s projects without issue or added complexity if there isn’t sensitive information.

ajansenn:
+second that

PR #1022, opened 2020-11-03
feat: Make security tokens optional

wbreza:
Security tokens are a great way to protect sensitive data from being stored in plain text within your project file, but at the same time can be more difficult to use when sharing projects with team members.

In this feature security tokens are still enabled by default but give the project author the ability to turn off security tokens which then stores all provider options in plain text.

Fixes: #835 #security

image


src/common/mockFactory.ts
```diff
@@ -277,6 +277,7 @@ export default class MockFactory {
             id: `project-${name}`,
             name: `Project ${name}`,
             version: appInfo.version,
+            useSecurityToken: true,
```
tessmichi, line 280: is there no mock for not using security token? if not, how is that tested?

tessmichi (approved):
lgtm

src/common/mockFactory.ts
wbreza: No, but i have a test that modifies the value after getting it from the mock.

tbarlow12 (approved):
ship it

jding05 (approved):

PR #1024, opened 2020-11-03
WIP - updated some dependencies

pjlittle:
- Updated to most recent version of `react-scripts` (4.0)
  - May still be an issue w/ referencing a specific version of `react-router` typings (5.0)
- Update TypeScript
- Fixed compilation errors 
- Migrated from tslint --> eslint
  - disabled a TON of warnings/errors (1000s)


fishcharlie:
@pjlittle @wbreza 

What is the status on this? With another Electron upgrade I think this should fix #1065 as well (https://www.electronjs.org/blog/apple-silicon/). Maybe also we can include another Electron update in here? Based on the TODO list in the original comment (and looking at the changes) it looks like the pipelines have been updated, but testing on Windows is the only thing that needs to be done?

I'd be happy to help dedicate some time to helping with this. Really want to get #1065 fixed. But there are some indications that this project has been abandoned. So don't want to sink a bunch of time into a project that is dead.

cc/ @tbarlow12 (since it looks like you have a lot of contributions to this project)

PR #1025, opened 2020-11-03
fix: Bing Image Search - Support to override endpoint url

wbreza:
The `Bing Search` resource is undergoing a transition from a `Cognitive Service` to the `Azure Marketplace`.  During this transition it is possible to have bing search resources that use different endpoint urls.

This fix introduces a new `endpoint` textbox that defaults to the new default resource url (https://api.bing.microsoft.com/).  However, for users with older resources that have not specified the field it will fall back to the older cognitive services url ( https://api.cognitive.microsoft.com/bing).

In addition to the fix this also introduces addition filter options to filter image responses by image size and license type.

TODO
- [x] Write new unit tests
- [x] Add localization

image

### DIFF
--- a/src/providers/storage/bingImageSearch.json
+++ b/src/providers/storage/bingImageSearch.json
@@ -18,7 +18,7 @@
         "query": {
             "type": "string",
             "title": "${strings.connections.providers.bing.query}",
-            "description": "The "
+            "description": "The Bing image search query"
         },
         "aspectRatio": {
             "type": "string",
--- a/src/providers/storage/bingImageSearch.json
+++ b/src/providers/storage/bingImageSearch.json
@@ -1,6 +1,6 @@
 {
     "type": "object",
-    "title": "${strings.connections.providers.bing.options}",
+    "title": "${strings.connections.providers.bing.options.title}",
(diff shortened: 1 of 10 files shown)


src/providers/storage/bingImageSearch.json
```diff
@@ -3,13 +3,22 @@
...
             "type": "string",
-            "title": "${strings.connections.providers.bing.apiKey}"
+            "title": "${strings.connections.providers.bing.apiKey}",
+            "description": "The key from your Bing Search Azure resource"
         },
         "query": {
             "type": "string",
-            "title": "${strings.connections.providers.bing.query}"
+            "title": "${strings.connections.providers.bing.query}",
+            "description": "The "
```
njm3754: Is this supposed to more that just "The"?

njm3754 (changes requested):

njm3754 (approved):

src/providers/storage/bingImageSearch.json
wbreza: Yes, fixed.

jding05 (approved):
LGTM

Issue #762, opened 2019-04-15
Feature: Ability to move project source location and maintain asset IDs


devedse:
I just recently moved the files I've been tagging with VOTT which resulted in me having to effectively write a tool to fix the .vott file.

Currently all the File ID's are generated based on an MD5 hash of the file path. However if you move these it will find all files a second time and then they are all marked as "not yet seen / tagged".

My suggestion would be is that file path's are being made relative to the .vott file. This would hopefully solve these issues.

wbreza:
@devedse - thanks for the feedback.  Using relative paths could work in the scenario and we will take a look into this.

cnrmck:
@devedse Could you share your tool for fixing the vott file? 

Would be nice to avoid some work/I'm sure others could use it.

cnrmck:
@devedse Just following up on this, having access to your tool would be really nice.

I really need the ability to modify the assets that someone else has labelled. 

I can drop their directory onto my computer, but when I open the .vott file (after `sed -i 's+/old/path/to/file+/new/path/to/file+g' /path/to/target/connection/directory/*`) I get an Unknown Error: 
![image](https://user-images.githubusercontent.com/22061955/57255595-8636bd00-7011-11e9-8b6a-38fdb11f0928.png)

I'm guessing that is due to my asset-ids not matching the hash of the absolute path. I'll write the script to fix this and share it here, but I second the importance of being able to share VoTT tags between machines (i.e. switching to relative paths to gen the asset-ids)

cnrmck:
Ok, here is a [python script](https://github.com/cnrmck/update-vott-assets) that will properly rename all of your asset files, references to asset ids, and paths. 

It also includes instructions on how to get your files to load in VoTT after running the script.

@devedse I'm not sure what I did differently from you without seeing your approach, but VoTT is successfully keeping track of my "not yet seen / tagged" status when I reopen the files!

devedse:
@cnrmck, my tool isn't really ready to be released, I'd suggest people use your script as it's well documented and probably does the same :)

devedse:
What version will this be added in? And is there also an upgrade scenario available to upgrade your old .vott file and asset.json files?

PR #1027, opened 2020-11-04
feat: Use relative paths for local file assets

tbarlow12:
All new connections will have a property `relativePath` that will indicate to use a path relative to the source connection when interacting with assets. Old connections will assume absolute path.

Addresses #762

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11916,6 +11916,12 @@
                 }
             }
         },
+        "mock-fs": {
+            "version": "4.13.0",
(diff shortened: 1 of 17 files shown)


jding05 (approved):

njm3754 (approved):

wbreza (approved):
Looks good 👍

lukasalexanderweber:
this would be a great improvement! Do you already know when this will be released?

Issue #839, opened 2019-07-28
Nested local directories


thepian:
**Is your feature request related to a problem? Please describe.**
A useful way to organise pictures is by organising them in folders named by the primary label

**Describe the solution you'd like**
Support a tree scan of source location rather than simple listdir

**Describe alternatives you've considered**
Alternately I have to create a connection per tag, which quickly becomes unwieldy, and it prevents exporting a single bundle.

Another option is to put all images in a single directory and name the images with the label, but I don't see why this should be the only option.

PR #1028, opened 2020-11-04
feat: Nested source directories

tbarlow12:
Adds capability of using all nested assets within a source directory.

Addresses #839

### DIFF
--- a/src/electron/providers/storage/localFileSystem.ts
+++ b/src/electron/providers/storage/localFileSystem.ts
@@ -93,17 +93,15 @@ export default class LocalFileSystem implements IStorageProvider {
         });
     }
 
-    public async listFiles(folderPath: string, ext?: string, recursive: boolean = false): Promise<string[]> {
+    public async listFiles(folderPath: string): Promise<string[]> {
         const normalizedPath = path.normalize(folderPath);
         console.log(`Listing files from ${normalizedPath}`);
         const files = await this.listItems(normalizedPath, (stats) => !stats.isDirectory());
-        if (recursive) {
-            const directories = await this.listItems(normalizedPath, (stats) => stats.isDirectory());
-            await directories.forEachAsync(async (directory) => {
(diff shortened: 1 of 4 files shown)


src/electron/providers/storage/localFileSystem.ts
```diff
@@ -137,9 +147,12 @@ export default class LocalFileSystem implements IStorageProvider {
         });
     }
 
-    public async getAssets(sourceConnectionFolderPath?: string, relativePath: boolean = false): Promise<IAsset[]> {
-        return (await this.listFiles(path.normalize(sourceConnectionFolderPath)))
-            .map((filePath) => AssetService.createAssetFromFilePath(
+    public async getAssets(
+            sourceConnectionFolderPath?: string,
+            relativePath: boolean = false,
+            recursive: boolean = true): Promise<IAsset[]> {
```
wbreza, line 153: Any reason to not always have `recursive` set??

src/electron/providers/storage/localFileSystem.ts
```diff
@@ -93,8 +93,18 @@ export default class LocalFileSystem implements IStorageProvider {
         });
     }
 
-    public listFiles(folderPath: string): Promise<string[]> {
-        return this.listItems(path.normalize(folderPath), (stats) => !stats.isDirectory());
+    public async listFiles(folderPath: string, ext?: string, recursive: boolean = false): Promise<string[]> {
```
wbreza, line 96: Same, any reason not do always have this as true?

wbreza (approved):
Looks great - Added some feedback on whether we should just always default to `recursive: true`?

src/electron/providers/storage/localFileSystem.ts
tbarlow12: Wasn't sure. Was leaving it open for flexibility, but I'll remove that param

jding05 (approved):
LGTM

PR #1051, opened 2021-04-12
fix: Adds missing registration of mixins to support 'forEachAsync'

wbreza:
The electron main process did not register the mixins which was causing recursive file system calls to fail.

- [x] Registers mixin @ electron main
- [x] Removes console.log statement


tbarlow12 (approved):
LGTM - good catch

Issue #980, opened 2020-06-24
Zoom functionality


damien-ls:
Working on high-resolution images where there is the need to annotate surface defects. The image is big and the defects are very small, invisible without zooming in.

Requested Functionality that allows to zoom-in/out with the mouse scroll (or equivalent).

Tried to use the windows magnifier, whilst it magnifies, it doesn't actually zoom in and the resolution is too poor to see details.

I believe this is a must-have feature, for us it is a straight killer that makes this application unusable for our use-case.

PR #1063, opened 2021-10-14
Zoom functionality by the mouse wheel

europeanplaice:
This solves #980.

I implemented zoom in/out by mouse wheel.
However, This is just a draft proposal, and I know there are a lot of problems for the time being.


giorgiococci:
Hi @europeanplaice,

I've tried your solution, the zoom works well, but the biggest problem that I found is that the two lines (horizontal and vertical) used to draw the bounding boxes are misaligned after zooming. Have you already found a solution for this problem? Thanks.

